PR #130, opened 2023-08-08
feat: store MM fade rate in dynamo

ConjunctiveNormalForm:
Also, fixed the SQL query that calculates fade rate - it was using archived orders table so only **filled orders** are considered


rileydcampbell (approved):
lgtm

PR #132, opened 2023-08-10
feat: change how we count faded quotes

ConjunctiveNormalForm:
Change the Redshift query to do the following instead:
- get the last 20 quotes from the last 7 days for each filler address
- count total and unfilled orders (`actualFiller` != original filler address in quote)
- filter out those who provided less than 10 quotes

Also, move fade rate calculation to quote handler (future PR) since a filler entity can have multiple addresses

### DIFF
--- a/lib/cron/fade-rate.ts
+++ b/lib/cron/fade-rate.ts
@@ -171,6 +171,10 @@ WITH latestOrders AS (
 SELECT
     latestOrders.chainid as chainId, latestOrders.filler as rfqFiller, latestOrders.quoteid, archivedorders.filler as actualFiller, latestOrders.createdat as postTimestamp, archivedorders.txhash as txHash,
     CASE
+      WHEN latestOrders.inputstartamount = latestOrders.inputendamount THEN 'EXACT_INPUT'
+      ELSE 'EXACT_OUTPUT'
+    END as tradeType, 
+    CASE
       WHEN latestOrders.inputstartamount = latestOrders.inputendamount THEN latestOrders.outputstartamount
       ELSE latestOrders.inputstartamount
     END as quotedAmount,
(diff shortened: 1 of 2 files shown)


lib/cron/fade-rate.ts
```diff
@@ -161,33 +161,48 @@ function createDynamoEntity() {
...
 `;
 
 const FADE_RATE_SQL = `
 WITH ORDERS_CTE AS (
     SELECT 
         rfqFiller,
-        COUNT(*) AS TotalFills,
-        SUM(CASE WHEN rfqFiller != actualFiller THEN 1 ELSE 0 END) AS UnmatchedFills
+        COUNT(*) AS totalQuotes,
+        SUM(CASE WHEN rfqFiller != actualFiller OR quotedAmount != filledAmount THEN 1 ELSE 0 END) AS fadedQuotes
```
blairmason, line 195: This will count sniped orders as fadedQuotes, which occurs on about 1% of orders. Not sure if that matters for this context. If so, I think you can honestly simplify it to just `quotedAmount > filledAmount`

lib/cron/fade-rate.ts
```diff
@@ -161,33 +161,52 @@ function createDynamoEntity() {
...
 `;
 
 const FADE_RATE_SQL = `
 WITH ORDERS_CTE AS (
     SELECT 
         rfqFiller,
-        COUNT(*) AS TotalFills,
-        SUM(CASE WHEN rfqFiller != actualFiller THEN 1 ELSE 0 END) AS UnmatchedFills
+        COUNT(*) AS totalQuotes,
+        SUM(CASE WHEN rfqFiller != actualFiller OR (tradeType = 'EXACT_INPUT' AND quotedAmount > filledAmount) OR (tradeType = 'EXACT_OUTPUT' AND quotedAmount < filledAmount) THEN 1 ELSE 0 END) AS fadedQuotes
```
blairmason, line 199: I think `rfqFiller != actualFiller OR ` should be removed. You will still count the overrides in this case.

rileydcampbell (approved):
don't have the sharpest SQL skills but this lgtm, deferring to Blair for SQL feedback :)

PR #137, opened 2023-08-25
Add SQL query for historical synthetic orders, dynamo integration

zhongeric:
tested locally e2e and it works, few things:
- need to tweak timeout + memory, especially as # of configs grow and when handling production quantity of data
- need to refactor if we change tokenConfig schema


lib/providers/token-price/index.ts
```diff
@@ -0,0 +1,82 @@
...
+        }
+        const response = await axios.post<any>(this.endpoint, payload, {
+            headers: {
+              'content-type': 'application/json',
+            },
+          })
+
+        return ethers.utils.parseUnits(response.data.quoteDecimals, token.decimals)
+    }
+}
```
ConjunctiveNormalForm, line 82:
```suggestion
}

```

lib/providers/token-price/index.ts
```diff
@@ -0,0 +1,82 @@
...
+            amount: ONE_USDC,
+            type: 'EXACT_OUTPUT',
+            configs: [
+              {
+                protocols: ['V2', 'V3', 'MIXED'],
+                routingType: 'CLASSIC',
+              },
+            ],
+        }
+        const response = await axios.post<any>(this.endpoint, payload, {
```
ConjunctiveNormalForm, line 74: this looks like an URA call? I was thinking we might be able to use a graphql endpoint from backend team which should be a lot cheaper

lib/providers/token-price/index.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export type BucketRange = {
+    lower: BigNumber;
+    upper: BigNumber;
+}
+
+export type TokenAmountsBucket = {
+    bucketRange: BucketRange 
+    lower: BigNumber;
+    upper: BigNumber;
```
ConjunctiveNormalForm, line 15: maybe call these `lowerInUSDC`, `upperInUSDC` for better clarity

ConjunctiveNormalForm (commented):
maybe change base branch from main to synth-pair-cron?

lib/providers/token-price/index.ts
zhongeric: yea down to use graphql especially if we imagine there to be a lot of token pairs in the config

lib/providers/token-price/index.ts
zhongeric: not doing the provider anymore

lib/cron/synth-switch.ts
```diff
@@ -41,14 +79,170 @@ const handler: ScheduledHandler = async (_event: EventBridgeEvent<string, void>)
...
+      status.Status === StatusString.SUBMITTED
+    ) {
+      await sleep(2000);
+    } else if (status.Status === StatusString.FINISHED) {
+      break;
+    } else {
+      log.error({ error: status.Error }, 'Unknown status');
+      throw new Error(status.Error);
+    }
+  }
```
rileydcampbell, line 226: nit: looks like this for loop code for waiting on the SQL command to return is repeated down below. Anyway we could turn this into some kind of function for reuse? no worries if that gets too tricky. I know there's some logic we are doing inside of it to so maybe not the cleanest thing to consolidate.

lib/cron/synth-switch.ts
```diff
@@ -108,9 +329,84 @@ async function readTokenConfig(log: Logger): Promise<TokenConfig[]> {
...
+      ...config,
+      inputToken: ethers.utils.getAddress(config.inputToken).toLowerCase(),
+      outputToken: ethers.utils.getAddress(config.outputToken).toLowerCase(),
+    };
+  });
+
+  return configs;
 }
 
+const CREATE_COMBINED_URA_RESPONSES_VIEW_SQL = `
```
rileydcampbell, line 350: chatGPT is a beast 🫡

lib/cron/synth-switch.ts
```diff
@@ -41,14 +79,170 @@ const handler: ScheduledHandler = async (_event: EventBridgeEvent<string, void>)
...
+            pos: 0,
+            neg: 0,
+          };
+        }
+        if (hasPriceImprovement) {
+          tradeOutcomesByKey[key].pos++;
+        } else {
+          tradeOutcomesByKey[key].neg++;
+        }
+      }
```
rileydcampbell, line 157: nit: might be nice to pull this logic out into another function something like `hasPositiveTradeOutcome()` then we can add any kind of arbitrary logic into that function to determine if trade was favorable. Mainly thinking about this if we wanted to add some kind of check for trade duration.

lib/providers/token-price/index.ts
```diff
@@ -0,0 +1,73 @@
...
+      tokenOutChainId: this.chainId,
+      amount: ONE_USDC,
+      type: 'EXACT_OUTPUT',
+      configs: [
+        {
+          protocols: ['V2', 'V3', 'MIXED'],
+          routingType: 'CLASSIC',
+        },
+      ],
+    };
```
rileydcampbell, line 62: Just thinking about this, if we know we want `CLASSIC` here maybe we should just call routing API directly? Saves a little bit of latency. Prob doesn't matter but at least we don't have to draw a circle if we ever diagramed this system out lol

lib/repositories/switch-repository.ts
```diff
@@ -84,4 +84,20 @@ export class SwitchRepository implements BaseSwitchRepository {
...
     console.log(JSON.stringify(res, null, 2));
   }
+
+  static getKey(trade: SynthSwitchTrade): string {
+    const { inputToken, inputTokenChainId, outputToken, outputTokenChainId, type } = trade;
+    return `${inputToken}#${inputTokenChainId}#${outputToken}#${outputTokenChainId}#${type}`;
+  }
+
+  static parseKey(key: string): SynthSwitchTrade {
+    const [inputToken, inputTokenChainId, outputToken, outputTokenChainId, type] = key.split('#');
```
rileydcampbell, line 94: nit: maybe throw with a meaningful error msg here if the `key.split` doesn't fit the format of `[inputToken, inputTokenChainId, outputToken, outputTokenChainId, type]`. Could just check key length ?

rileydcampbell (commented):
Overall lgtm, had some comments but nothing major. After that and some tests prob good to ship.

lib/cron/synth-switch.ts
```diff
@@ -22,6 +28,28 @@ type TokenConfig = {
...
+  tokenout: string;
+  tokenoutchainid: string;
+  dutch_amountout: string;
+  classic_amountout: string;
+  dutch_amountoutgasadjusted: string;
+  classic_amountoutgasadjusted: string;
+  settledAmountIn: string;
+  settledAmountOut: string;
+  filler: string;
+  filltimestamp: string;
```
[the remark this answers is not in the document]
zhongeric: hmm I could camel case, the ones with dutch_ and classic_ prefixed are just how I wrote the sql statement

lib/cron/synth-switch.ts
zhongeric: yo this was all me man

lib/cron/synth-switch.ts
```diff
@@ -41,14 +79,170 @@ const handler: ScheduledHandler = async (_event: EventBridgeEvent<string, void>)
...
+      // build trade objects differentiating between ExactIn and ExactOut
+      let tradeOutcomesByKey: {
+        [key: string]: {
+          pos: number;
+          neg: number;
+        };
+      } = {};
+      for (const order of ordersForConfig) {
+        const tradeType =
+          order.classic_amountin == order.classic_amountingasadjusted ? TradeType.EXACT_INPUT : TradeType.EXACT_OUTPUT;
```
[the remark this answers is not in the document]
rileydcampbell: actually scratch this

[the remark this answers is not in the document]
zhongeric: yeah this was a kind of weird method but we can look at where gas is taken out of to tell tradetype

lib/cron/synth-switch.ts
```diff
@@ -41,14 +79,170 @@ const handler: ScheduledHandler = async (_event: EventBridgeEvent<string, void>)
...
+        const trade: SynthSwitchRequestBody = {
+          inputToken: order.tokenin,
+          inputTokenChainId: Number(order.tokeninchainid),
```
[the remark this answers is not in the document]
zhongeric: sg, changing to parseint

lib/cron/synth-switch.ts
```diff
@@ -41,14 +79,170 @@ const handler: ScheduledHandler = async (_event: EventBridgeEvent<string, void>)
...
+          outputToken: order.tokenout,
+          outputTokenChainId: Number(order.tokenoutchainid),
+          type: String(tradeType),
+          // classic amount in here or synthetic amount in?
```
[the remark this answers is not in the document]
zhongeric: yeah idt it matters

lib/cron/synth-switch.ts
```diff
@@ -41,14 +79,170 @@ const handler: ScheduledHandler = async (_event: EventBridgeEvent<string, void>)
...
+          tradeOutcomesByKey[key].neg++;
+        }
+      }
 
+      Object.keys(tradeOutcomesByKey).forEach(async (key) => {
+        const { pos, neg } = tradeOutcomesByKey[key];
+        const totalOrders = pos + neg;
+        if (totalOrders >= MINIMUM_ORDERS) {
+          if (neg / totalOrders >= DISABLE_THRESHOLD) {
+            // TODO: update tradeSizes with new TokenConfig schema
```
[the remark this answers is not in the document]
zhongeric: this TODO was mainly for me tmrw to update it with the new token config schema from zach, but yeah thinking through it right now the logic will ignore trade size and turn off for all token pairs.

We could do something where we set it to true still but set the lower amount as the largest amount in all of the orders?

lib/cron/synth-switch.ts
zhongeric: oo yeah good idea

lib/providers/token-price/index.ts
zhongeric: Yeah sounds good, I'm thinking of just removing these files from the PR for now since we're not shipping it in the first version

lib/repositories/switch-repository.ts
zhongeric: hm length is hard because chainIds can be different lengths

zhongeric: I can check for definedness though at least

lib/cron/synth-switch.ts
```diff
@@ -41,14 +79,178 @@ const handler: ScheduledHandler = async (_event: EventBridgeEvent<string, void>)
...
+      hasPriceImprovement = BigNumber.from(order.settledAmountOut).gt(order.classic_amountoutgasadjusted);
+    } else {
+      hasPriceImprovement = BigNumber.from(order.classic_amountingasadjusted).gt(order.settledAmountIn);
+    }
+    // can add more conditionals here
+    const result = hasPriceImprovement;
+    return { key, result };
+  }
+
+  async function updateSynthSwitchRepository(configs: TokenConfig[], result: ResultRowType[]) {
```
ConjunctiveNormalForm, line 125: if we do all the calculations in lambda sequentially, which I think is fine for a cron job, we should give it enough memory/timeout threshold - not sure if current setting needs a change tho justing calling out

lib/cron/synth-switch.ts
```diff
@@ -41,14 +79,178 @@ const handler: ScheduledHandler = async (_event: EventBridgeEvent<string, void>)
...
+              {
+                key,
+                ordersWithNegativeOutcome: neg,
+                ordersWithPositiveOutcome: pos,
+                totalOrders,
+              },
+              'Disabling synthethics for trade'
+            );
+            // TODO: update tradeSizes with new TokenConfig schema
+            await synthSwitchEntity.putSynthSwitch(SwitchRepository.parseKey(key), config.tradeSizes[0], false);
```
ConjunctiveNormalForm, line 182: trying to think through a potential edge case: say we originally have { in: USDC, inChain: 1, out: UNI, outChain: 1, type: ExactIn, lower: 100,000 } in the config and enabled = false. If somehow we lower `lower` to 10,000 and the cron decided to put `enabled = true` we'd then have two rows with the same partition key. Currently we query the table in reverse order meaning the row with `enabled = false`  will be returned.

maybe it makes more sense to remove the sort key (lower) from the table? so only one row (or none) can exist

zhongeric: > so only one row (or none) can exist
i was under the impression that this was true in the current implementation but yeah I see what you mean. I thought that the default behavior if the partition key exists is that it would update it with the new sort key value?

lib/cron/synth-switch.ts
zhongeric: smart yeah, increasing timeout to 10 min and will reduce after testing

lib/cron/synth-switch.ts
zhongeric: hm yeah could do something like `handleStmtResult`, down for a fast follow on this? like you said I think we have some logic in the FINISHED state for one use case

ConjunctiveNormalForm (approved):
lgtm

PR #138, opened 2023-08-28
feat: add dynamo repository

ConjunctiveNormalForm:


lib/cron/synth-switch.ts
```diff
@@ -16,6 +16,8 @@ import { checkDefined } from '../preconditions/preconditions';
 type TokenConfig = {
   inputToken: string;
   outputToken: string;
+  inputTokenChainId: number;
+  outputTokenChainId: number;
```
ConjunctiveNormalForm, line 20: I think we were missing chainIds here and also in the config?

lib/repositories/switch-repository.ts
```diff
@@ -0,0 +1,81 @@
+import { DynamoDBDocumentClient } from '@aws-sdk/lib-dynamodb';
+import Logger from 'bunyan';
+import { Entity, Table } from 'dynamodb-toolbox';
+
+import { DYNAMO_TABLE_KEY, DYNAMO_TABLE_NAME } from '../constants';
+import { SynthSwitchRequestBody } from '../handlers/synth-switch';
+import { BaseSwitchRepository } from './base';
+
+export const PARTITION_KEY = `${DYNAMO_TABLE_KEY.INPUT_TOKEN}#${DYNAMO_TABLE_KEY.INPUT_TOKEN_CHAIN_ID}${DYNAMO_TABLE_KEY.OUTPUT_TOKEN}#${DYNAMO_TABLE_KEY.OUTPUT_TOKEN_CHAIN_ID}#${DYNAMO_TABLE_KEY.TRADE_TYPE}`;
```
ConjunctiveNormalForm, line 9: partition key + sort key must uniquely identify an item in dynamo

lib/repositories/switch-repository.ts
```diff
@@ -0,0 +1,81 @@
...
+    const { inputToken, inputTokenChainId, outputToken, outputTokenChainId, type, amount } = trade;
+
+    // get row for which lower bucket <= amount < upper bucket
+    const result = await this.switchEntity.query(
+      `${inputToken}#${inputTokenChainId}#${outputToken}#${outputTokenChainId}#${type}`,
+      {
+        limit: 1,
+        filters: [
+          { attr: 'lower', lte: amount },
+          { attr: 'upper', gt: amount },
```
ConjunctiveNormalForm, line 56: I imagine URA will call the `enabled` endpoint with native token amount (tokenIn amount if exact_input, etc), where it will be checked against the lower and upper bucket values, also in native amount, that are stored in our dynamo table 
@zhongeric let me know if that makes sense to you

zhongeric: yeah this makes sense, URA should be calling with the lower and upper amounts in terms of the token, not USD. We should have the bucket logic in GPA for sure

lib/repositories/base.ts
```diff
@@ -0,0 +1,6 @@
+import { SynthSwitchRequestBody } from '../handlers/synth-switch';
```
zhongeric, line 1: this file doesn't exist btw on this branch

lib/constants.ts
```diff
@@ -11,8 +11,11 @@ export const DYNAMO_TABLE_NAME = {
...
   FILLER: 'filler',
   INPUT_TOKEN: 'inputToken',
-  OUTPUT_TKEN: 'outputToken',
-  TRADE_TYPE: 'tradeType',
-  TRADE_SIZE: 'tradeSize',
+  INPUT_TOKEN_CHAIN_ID: 'inputTokenChainId',
+  OUTPUT_TOKEN: 'outputToken',
+  OUTPUT_TOKEN_CHAIN_ID: 'outputTokenChainId',
+  TRADE_TYPE: 'type',
+  UPPER: 'upper',
```
zhongeric, line 18: removing UPPER right?

lib/repositories/switch-repository.ts
```diff
@@ -0,0 +1,81 @@
...
```
zhongeric, line 56: removing upper

lib/repositories/switch-repository.ts
```diff
@@ -0,0 +1,81 @@
...
+    lower: string,
+    upper: string,
+    enabled: boolean
+  ): Promise<void> {
+    const { inputToken, inputTokenChainId, outputToken, outputTokenChainId, type } = trade;
+
+    await this.switchEntity.put({
+      [PARTITION_KEY]: `${inputToken}#${inputTokenChainId}#${outputToken}#${outputTokenChainId}#${type}`,
+      [`${DYNAMO_TABLE_KEY.LOWER}`]: lower,
+      [`${DYNAMO_TABLE_KEY.UPPER}`]: upper,
```
zhongeric, line 77: .

zhongeric (approved):

lib/handlers/synth-switch/schema.ts
```diff
@@ -0,0 +1,29 @@
+import Joi from 'joi';
+
+import { FieldValidator } from '../../util/validator';
+
+export const SynthSwitchRequestBodyJoi = Joi.object({
+  tokenInChainId: FieldValidator.chainId.required(),
+  inputTokenChainId: Joi.number().integer().valid(Joi.ref('tokenInChainId')).required(),
```
rileydcampbell, line 7: Assume we don't need both of these?

lib/repositories/switch-repository.ts
```diff
@@ -0,0 +1,78 @@
...
+    } as const);
+
+    return new SwitchRepository(switchTable, switchEntity);
+  }
+
+  private constructor(
+    // eslint-disable-next-line
+    private readonly _switchTable: Table<
+      'SynthSwitch',
+      'inputToken#inputTokenChainId#outputToken#outputTokenChainId#type',
```
rileydcampbell, line 45:
```suggestion
      PARTITION_KEY,
```
maybe?

lib/repositories/switch-repository.ts
```diff
@@ -0,0 +1,78 @@
...
+  ) {}
+
+  public async syntheticQuoteForTradeEnabled(trade: SynthSwitchRequestBody): Promise<boolean> {
+    const { inputToken, inputTokenChainId, outputToken, outputTokenChainId, type, amount } = trade;
+
+    // get row for which lower bucket <= amount
+    const result = await this.switchEntity.query(
+      `${inputToken}#${inputTokenChainId}#${outputToken}#${outputTokenChainId}#${type}`,
+      {
+        limit: 1,
```
rileydcampbell, line 58: If there are multiple rows that the lower is lte `amount` does this return the biggest one?

test/repositories/switch-repository.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  },
+  unmarshallOptions: {
+    wrapNumbers: true,
+  },
+});
+
+const switchRepository = SwitchRepository.create(documentClient);
+
+describe('put switch tests', () => {
+  it('should put synth switch into db', async () => {
```
rileydcampbell, line 48: nit: should we add a unit test for the case when we have multiple trade sizes of the same swap in the DB?

rileydcampbell (commented):
overall lgtm, just couple nits

lib/repositories/switch-repository.ts
ConjunctiveNormalForm: yeah unfortunately PARTITION_KEY is a value so cannot be used as a type parameter here (at least that's what the linter says)

lib/repositories/switch-repository.ts
ConjunctiveNormalForm: think in our design at most one row exists for a specific input/output/chainIds/type combo

PR #140, opened 2023-08-30
feat: add synth-switch lambda

ConjunctiveNormalForm:
add `/synthetic-swtich/enabled` endpoint. 

will add some basic handler tests
Screenshot 2023-08-30 at 11 16 14 PM

### DIFF
--- a/bin/stacks/api-stack.ts
+++ b/bin/stacks/api-stack.ts
@@ -316,7 +316,18 @@ export class APIStack extends cdk.Stack {
       },
     });
     const enabled = switchResource.addResource('enabled');
-    enabled.addMethod('GET', switchLambdaIntegration);
+
+    /* add auth key */
+    const apiAuthzKey = api.addApiKey('AuthzKey');
+    const plan = api.addUsagePlan('AccessPlan', {
+      name: 'AccessPlan',
+    });
+    plan.addApiKey(apiAuthzKey);
+    plan.addApiStage({
+      stage: api.deploymentStage,
+    });
+
+    enabled.addMethod('GET', switchLambdaIntegration, { apiKeyRequired: true });
 
     const integration = api.root.addResource('integration', {
       defaultCorsPreflightOptions: {
--- a/bin/stacks/api-stack.ts
+++ b/bin/stacks/api-stack.ts
@@ -209,6 +209,31 @@ export class APIStack extends cdk.Stack {
       provisionedConcurrentExecutions: provisionedConcurrency > 0 ? provisionedConcurrency : undefined,
     });
 
+    const switchLambda = new aws_lambda_nodejs.NodejsFunction(this, 'Switch', {
+      role: lambdaRole,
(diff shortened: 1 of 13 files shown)


bin/stacks/api-stack.ts
```diff
@@ -283,6 +308,16 @@ export class APIStack extends cdk.Stack {
     });
     quote.addMethod('POST', quoteLambdaIntegration);
 
+    const switchLambdaIntegration = new aws_apigateway.LambdaIntegration(switchLambdaAlias, {});
+    const switchResource = api.root.addResource('synthetic-switch', {
+      defaultCorsPreflightOptions: {
+        allowOrigins: aws_apigateway.Cors.ALL_ORIGINS,
+        allowMethods: aws_apigateway.Cors.ALL_METHODS,
```
zhongeric, line 315: we should only allow URA to call this endpoint right? or else people will be able to sniff out whats enabled / not

zhongeric: not CORS related but we can setup an internal api key
=

ConjunctiveNormalForm: added auth key and made it required for `enabled` endpoint

bin/stacks/api-stack.ts
```diff
@@ -316,7 +316,18 @@ export class APIStack extends cdk.Stack {
       },
     });
     const enabled = switchResource.addResource('enabled');
-    enabled.addMethod('GET', switchLambdaIntegration);
+
+    /* add auth key */
+    const apiAuthzKey = api.addApiKey('AuthzKey');
```
zhongeric, line 321: where is this value being generated / fetched? I was envisioning a pattern more similar to how we do the GPA internal key for URA

ConjunctiveNormalForm: key is generated right here! We store it in pipeline secret manager and URA reads it and include it in x-api-key header when calling `enabled`

zhongeric: oh cool, is Authz a typo or related to this?

lib/repositories/switch-repository.ts
```diff
@@ -48,32 +47,29 @@ export class SwitchRepository implements BaseSwitchRepository {
     private readonly switchEntity: Entity
   ) {}
 
-  public async syntheticQuoteForTradeEnabled(trade: SynthSwitchRequestBody): Promise<boolean> {
+  public async syntheticQuoteForTradeEnabled(trade: SynthSwitchQueryParams): Promise<boolean> {
     const { inputToken, inputTokenChainId, outputToken, outputTokenChainId, type, amount } = trade;
 
+    SwitchRepository.log.info({ trade: trade });
```
zhongeric, line 53: this.log

zhongeric (approved):

lib/repositories/switch-repository.ts
ConjunctiveNormalForm: log is a static member here

PR #157, opened 2023-09-06
fix: synth lower comparison

zhongeric:


lib/repositories/switch-repository.ts
```diff
@@ -61,7 +62,7 @@ export class SwitchRepository implements BaseSwitchRepository {
     );
 
     SwitchRepository.log.info({ res: result.Item }, 'get result');
-    if (result.Item && result.Item.lower <= amount) {
+    if (result.Item && BigNumber.from(result.Item.lower).lte(amount)) {
```
rileydcampbell, line 65:
```suggestion
    if (result.Item && BigNumber.from(result.Item.lower).lte(BigNumber.from(amount))) {
```
? because amount is type `string` ?

rileydcampbell: oh actually looking at your test guess we don't need it

rileydcampbell (approved):

marktoda (approved):

PR #172, opened 2023-09-11
chore: set quoteId equal to requestId

rileydcampbell:
## Summary
Currently market makers have difficulty associating their quote response with the actual signed order. This is because we generate a random id for the `quoteId` returned from the param service. We send them the `requestId` of the param api request. This PR sets the `quoteId == requestId` so that you can directly link a RFQ quote request to the signed order.

Data concerns brought up in [this thread](https://uniswapteam.slack.com/archives/C052XQ3EUTT/p1694468331011049).

### DIFF
--- a/lib/entities/QuoteResponse.ts
+++ b/lib/entities/QuoteResponse.ts
@@ -1,11 +1,10 @@
 import { TradeType } from '@uniswap/sdk-core';
 import { BigNumber } from 'ethers';
 import { ValidationResult } from 'joi';
-import { v4 as uuidv4 } from 'uuid';
 
-import { QuoteRequestData } from '.';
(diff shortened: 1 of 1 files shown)


ConjunctiveNormalForm (approved):
Makes sense!

rileydcampbell:
Chatted with Alex about this change and we are gonna see what MMs think about receiving a new field called `quoteId` that is the current quote id we randomly generate. That way we aren't changing the data source of an existing field.

rileydcampbell:
Chatted with Alex about this change and we are gonna see what MMs think about receiving a new field called quoteId that is the current quote id we randomly generate. That way we aren't changing the data source of an existing field.

PR #182, opened 2023-09-15
fix: remove where in clause from sql query

zhongeric:


lib/cron/synth-switch.ts
```diff
@@ -398,8 +391,17 @@ async function main(metricsLogger: MetricsLogger) {
...
         return formattedRow;
       });
-      metrics.putMetric(Metric.SYNTH_ORDERS, formattedResult.length, MetricLoggerUnit.Count);
-      await updateSynthSwitchRepository(configs, formattedResult, metrics);
+
+      // filter out orders that are not in the configs
+      const filteredResult = formattedResult.filter((row) => {
+        return (
+          tokenInListRaw.includes(row.tokenin.toLowerCase()) &&
+          tokenOutListRaw.includes(row.tokenout.toLowerCase())
```
ConjunctiveNormalForm, line 399: if config has two entries A->B, C->D, then tokenInList is [A, C] and tokenOutList is [B, D], this would allow C->D trade?

zhongeric: ah yes good catch

ConjunctiveNormalForm (approved):
think a set with composite key would be a little more efficient? I'm fine with this tho

PR #186, opened 2023-09-18
feat: complete fade rate cron

ConjunctiveNormalForm:
Aggregating faded vs. filled quotes across multiple filler addresses by filler entities and store the result to S3 (which I think is more suitable than dynamo)

Also some refactoring of the repositories - main reason is to make it easier for different Redsfhit-base repositories to share a few methods. I did not touch synth-switch just to be extra safe


lib/repositories/base.ts
```diff
@@ -1,18 +1,66 @@
...
 
+export enum TimestampThreshold {
+  ONE_MONTH = "'1 MONTH'",
+  TWO_MONTHS = "'2 MONTHS'",
+}
+
+export abstract class BaseRedshiftRepository {
+  constructor(readonly client: RedshiftDataClient, private readonly configs: SharedConfigs) {}
+
+  async executeStatement(sql: string, log: Logger, executionConfigs?: ExecutionConfigs): Promise<string> {
```
rileydcampbell, line 33: nice moving this into the base class ++

rileydcampbell (approved):
lgtm

PR #187, opened 2023-09-19
feat: quoter calls only to eligible filler endpoints (fade rate < threshold)

ConjunctiveNormalForm:


lib/repositories/fades-repository.ts
```diff
@@ -62,7 +62,8 @@ WITH latestOrders AS (
   SELECT * FROM (
     SELECT *, ROW_NUMBER() OVER (PARTITION BY filler ORDER BY createdat DESC) AS row_num FROM postedorders
   )
-  WHERE row_num <= 20
+  WHERE row_num <= 30
```
zhongeric, line 65: would this be the last 30 orders or the last 30 orders per filler? if the former then could it be possible that there are orders from fillers not captured in this window?

test/providers/quoters/WebhookQuoter.test.ts
```diff
@@ -65,6 +75,67 @@ describe('WebhookQuoter tests', () => {
     expect(response[0].toResponseJSON()).toEqual({ ...quote, quoteId: expect.any(String) });
   });
 
+  it('Only calls to eligible endpoints', async () => {
```
zhongeric, line 78: nit: good to also put a comment here with what the threshold is for the breaker, and say which endpoints should be called / not called

zhongeric (approved):

lib/repositories/fades-repository.ts
ConjunctiveNormalForm: this is last 30 orders per filler

lib/quoters/WebhookQuoter.ts
```diff
@@ -30,6 +35,25 @@ export class WebhookQuoter implements Quoter {
     return QuoterType.RFQ;
   }
 
+  private async getEligibleEndpoints(): Promise<WebhookConfiguration[]> {
+    const endpoints = await this.webhookProvider.getEndpoints();
+    const config = await this.circuitBreakerProvider.getConfigurations();
```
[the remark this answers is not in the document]
ConjunctiveNormalForm: good call, added 500ms timeout to the S3 CircuitBreakerConfigProvider

test/providers/quoters/WebhookQuoter.test.ts
ConjunctiveNormalForm: here we actually uses the `enabled` field in the config, the fade rate is there just for reference

package.json
```diff
@@ -59,6 +59,7 @@
     "@aws-sdk/client-redshift-data": "^3.382.0",
     "@aws-sdk/client-s3": "^3.304.0",
     "@aws-sdk/lib-dynamodb": "^3.386.0",
+    "@smithy/node-http-handler": "^2.1.5",
```
zhongeric, line 62: why the new dep?

lib/quoters/WebhookQuoter.ts
```diff
@@ -31,6 +39,32 @@ export class WebhookQuoter implements Quoter {
...
+  private async getEligibleEndpoints(): Promise<WebhookConfiguration[]> {
+    const endpoints = await this.webhookProvider.getEndpoints();
+    try {
+      const config = await this.circuitBreakerProvider.getConfigurations();
+      const fillerToConfigMap = new Map(config.map((c) => [c.name, c]));
+      if (config) {
+        const enabledEndpoints: WebhookConfiguration[] = [];
+        endpoints.forEach((e) => {
+          if (
+            this.ALLOW_LIST.has(e.name) ||
```
zhongeric, line 51: wouldn't we still need a way to change the dynamo table even with this allowlist setting? In order to turn the endpoints back on, or else everything will have to be in ALLOW_LIST eventually?

zhongeric (approved):

package.json
ConjunctiveNormalForm: to use the NodeHttpHandler for custom timeout
```
    this.client = new S3Client({
      requestHandler: new NodeHttpHandler({
        requestTimeout: 500,
      }),
    });
```

lib/quoters/WebhookQuoter.ts
ConjunctiveNormalForm: hmmmm you are right, we need a way to retire entries in the config, maybe a timestamp

the allow list is still needed tho otherwise they cron would just turn someone off again right after we re-enable

PR #192, opened 2023-09-20
fix: add stage field to lambda env

ConjunctiveNormalForm:
synth-switch cron lambda is trying to read from `synth-config-undefined-1` because `stage` is not in process env


rileydcampbell (approved):

bin/stacks/cron-stack.ts
```diff
@@ -94,6 +94,7 @@ export class CronStack extends cdk.NestedStack {
         REDSHIFT_DATABASE: RsDatabase,
         REDSHIFT_CLUSTER_IDENTIFIER: RsClusterIdentifier,
         REDSHIFT_SECRET_ARN: RedshiftCredSecretArn,
+        stage: stage,
```
rileydcampbell, line 97: Ah I think this is because in your your last PR we moved stage out of envVars but never passed it directly to lambda env like this.

PR #198, opened 2023-09-26
chore: add support for beta s3 config

rileydcampbell:
## Summary
In order to support our beta test env we need to change the s3 bucket we read from based on the stage we are deployed in.

### DIFF
--- a/lib/cron/fade-rate.ts
+++ b/lib/cron/fade-rate.ts
@@ -29,8 +29,9 @@ async function main(metrics: MetricsLogger) {
     name: 'FadeRate',
     serializers: Logger.stdSerializers,
   });
-  const s3Key = process.env['stage'] === STAGE.BETA ? BETA_S3_KEY : PRODUCTION_S3_KEY;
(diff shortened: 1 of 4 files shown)


lib/cron/fade-rate.ts
```diff
@@ -22,7 +29,8 @@ async function main(metrics: MetricsLogger) {
     name: 'FadeRate',
     serializers: Logger.stdSerializers,
   });
-  const webhookProvider = new S3WebhookConfigurationProvider(log, `${WEBHOOK_CONFIG_BUCKET}-prod-1`, PRODUCTION_S3_KEY);
+  const s3Key = process.env['stage'] === STAGE.BETA ? BETA_S3_KEY : PRODUCTION_S3_KEY;
+  const webhookProvider = new S3WebhookConfigurationProvider(log, `${WEBHOOK_CONFIG_BUCKET}-prod-1`, s3Key);
```
ConjunctiveNormalForm, line 33: nit: `${WEBHOOK_CONFIG_BUCKET}-${stage}-1`? but at the same time we disabled fade rate cron for beta, so really it doesn't matter

ConjunctiveNormalForm (approved):

PR #202, opened 2023-09-29
chore: cron tweaks

ConjunctiveNormalForm:
- Reduce Redshift retention to 1 month 
- delete (almost) all rows marked for deletion

### DIFF
--- a/lib/cron/redshift-reaper.ts
+++ b/lib/cron/redshift-reaper.ts
@@ -9,7 +9,6 @@ const TABLES_TO_CLEAN = [
   'unifiedroutingresponses',
   'rfqrequests',
   'rfqresponses',
-  'archivedorders',
   'postedorders',
 ];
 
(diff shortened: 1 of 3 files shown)


lib/cron/redshift-reaper.ts
```diff
@@ -9,7 +9,6 @@ const TABLES_TO_CLEAN = [
   'unifiedroutingresponses',
   'rfqrequests',
   'rfqresponses',
-  'archivedorders',
```
zhongeric, line 12: are you removing archivedOrders from this so we don't delete those?

zhongeric (approved):

lib/cron/redshift-reaper.ts
ConjunctiveNormalForm: realized archivedOrders don't have a timestamp field

PR #214, opened 2023-10-18
fix: use timestamps to calc fade rate

ConjunctiveNormalForm:


lib/repositories/fades-repository.ts
```diff
@@ -93,7 +88,7 @@ WITH ORDERS_CTE AS (
     SELECT 
         rfqFiller,
         COUNT(*) AS totalQuotes,
-        SUM(CASE WHEN (tradeType = 'EXACT_INPUT' AND quotedAmount > filledAmount) OR (tradeType = 'EXACT_OUTPUT' AND quotedAmount < filledAmount) THEN 1 ELSE 0 END) AS fadedQuotes
+        SUM(CASE WHEN (decayStartTime < fillTimestamp) THEN 1 ELSE 0 END) AS fadedQuotes
```
ConjunctiveNormalForm, line 91: don't have to check filler addr - if overridden within exclusivity window by someone else, we don't count that as fade

marktoda (approved):

PR #217, opened 2023-10-24
fix: use hash instead

ConjunctiveNormalForm:


lib/quoters/WebhookQuoter.ts
```diff
@@ -22,7 +22,7 @@ export class WebhookQuoter implements Quoter {
     _log: Logger,
     private webhookProvider: WebhookConfigurationProvider,
     private circuitBreakerProvider: CircuitBreakerConfigurationProvider,
-    _allow_list: Set<string> = new Set<string>(['Altonomy'])
+    _allow_list: Set<string> = new Set<string>(['9de8f2376fef4be567f2e242fce750cca347b71853816cbc64f70d568de41ef1'])
```
rileydcampbell, line 25: Would add Inu Paws too

rileydcampbell (approved):

PR #226, opened 2023-11-06
chore: log opposing quote iff real quote is valid

ConjunctiveNormalForm:


lib/quoters/WebhookQuoter.ts
```diff
@@ -170,6 +170,15 @@ export class WebhookQuoter implements Quoter {
           request.requestId
         } for endpoint ${endpoint} successful quote: ${request.amount.toString()} -> ${quote.toString()}}`
       );
+      
+      //iff valid quote, log the opposing side as well
+      const opposingRequest = request.toOpposingRequest();
+      const opposingResponse = QuoteResponse.fromRFQ(opposingRequest, opposite.data, opposingRequest.type).response;
+      this.log.info({
```
rileydcampbell, line 177: Should there be a condition here to log `if (opposingResponse)` ?

rileydcampbell (approved):

PR #228, opened 2023-11-08
feat: add filler compliance provider

ConjunctiveNormalForm:


lib/providers/compliance/mock.ts
```diff
@@ -0,0 +1,24 @@
+import { FillerComplianceConfiguration, FillerComplianceConfigurationProvider } from '.';
+
+export class MockFillerComplianceConfigurationProvider implements FillerComplianceConfigurationProvider {
+  constructor(private configs: FillerComplianceConfiguration[]) {}
+
+  async getConfigs(): Promise<FillerComplianceConfiguration[]> {
+    return this.configs;
+  }
+
+  async getAddrToEndpointsMap(): Promise<Map<string, Set<string>>> {
```
marktoda, line 10: maybe changing the name to be more clear whether this is endpoints to _include_ or endpoints to _exclude_ the address would be helpful for readability. Maybe `getExcludedEndpointsByAddr`

lib/providers/compliance/s3.ts
```diff
@@ -0,0 +1,60 @@
...
+    this.configs = [];
+    this.log = _log.child({ quoter: 'S3FillerComplianceConfigurationProvider' });
+    this.addrToEndpointsMap = new Map<string, Set<string>>();
+  }
+
+  async getAddrToEndpointsMap(): Promise<Map<string, Set<string>>> {
+    if (this.configs.length === 0) {
+      await this.fetchConfigs();
+    }
+    if (this.addrToEndpointsMap.size === 0) {
```
marktoda, line 24: nit: think early return is slightly more readable than conditioned setup
```
if (this.addrToEndpointsMap.size > 0) return this.addrToEndpoitnsMap
```

lib/providers/compliance/s3.ts
```diff
@@ -0,0 +1,60 @@
...
+  private addrToEndpointsMap: Map<string, Set<string>>;
+
+  constructor(_log: Logger, private bucket: string, private key: string) {
+    this.configs = [];
+    this.log = _log.child({ quoter: 'S3FillerComplianceConfigurationProvider' });
+    this.addrToEndpointsMap = new Map<string, Set<string>>();
+  }
+
+  async getAddrToEndpointsMap(): Promise<Map<string, Set<string>>> {
+    if (this.configs.length === 0) {
```
marktoda, line 21: do we want to refetch configs on a timer to capture updates without a redeploy? Or maybe lambda container cycling does this for us

lib/quoters/WebhookQuoter.ts
```diff
@@ -30,6 +32,19 @@ export class WebhookQuoter implements Quoter {
 
   public async quote(request: QuoteRequest): Promise<QuoteResponse[]> {
     const endpoints = await this.getEligibleEndpoints();
+    const addrToEndpointsMap = await this.complianceProvider.getAddrToEndpointsMap();
+    
+    endpoints.forEach((endpoint) => {
```
marktoda, line 37: could this instead just be a filter? splicing while iterating is a bit of a code smell

marktoda (approved):

lib/providers/compliance/s3.ts
ConjunctiveNormalForm: Yeah I decided to not refetch on our own because 1) the config doesn't update as often, definitely not every 15 minutes 2) it's much larger than other configs we have so I'd rather we only read it once during lambda cold start

PR #236, opened 2023-11-16
[feat] add WebhookQuoterResponse analytics event logging

blairmason:
## Background
Additional analytics events are needed for analyzing and reporting of quoter endpoint raw webhook responses. The initial implementation of logging analytics events was the following:
1. Log event to CloudWatch using Bunyan, specifying an `eventType` and putting the event properties as a nested objected in the `body` property.
2. Setting up a CloudWatch subscription filter that listens to the log group of a specific lambda function, then puts the matched logs to a Kinesis Data Firehose delivery stream specific to that event.
3. The Firehose stream processes the log records received from the subscription filter using a lambda function processor and then outputs to a dedicated S3 bucket for that `eventType`.
4. The Firehose delivery stream then copies the data from S3 and loads to a Redshift table.

This process has worked fine so far, but has the following limitations:
- Relies on 1 CloudWatch subscription filter and those have a resource cap of 2 per log group (per lambda function essentially)
- Optimizes for Redshift destination, but loading to our Google BigQuery data warehouse requires tracking many different S3 buckets and keeping track of the schema of each Redshift table.

## Changes
This PR introduces the following alternative lightweight analytics stack:

1. `lib/stacks/firehose-stack.ts` sets up a single Firehose delivery stream that can be reused by all handlers to log analytics events to the same destination S3 bucket as GZIP compressed newline JSON
2. `lib/providers/analytics/index.ts` defines `IAnalyticsLogger` interface
3. `lib/repositories/firehose-repository.ts` defines the `FirehoseLogger` class, and implementation of `IAnalyticsLogger`,  that facilitates sending analytics events to the analytics firehose destination stream using the `sendAnalyticsEvent()` method
4. `lib/entities/analytics-events.ts` defines a AnalyticsEvent class, which standardizes base object properties:
    - `eventType` - enum 
    - `eventTime` - set by constructor
    - `eventProperties` - semi-structured JSON object

This PR also replaces current CloudWatch logging for raw webhook responses with `WebhookQuoterResponse` analytics events using this new analytics stack and updates existing tests to validate logging.


lib/quoters/WebhookQuoter.ts
```diff
@@ -119,68 +132,71 @@ export class WebhookQuoter implements Quoter {
...
+        eventProperties: {
+          ...requestContext,
+          ...rawResponse,
+          responseType: WebhookResponseType.OK,
         },
-        `WebhookQuoter: request ${
-          request.requestId
-        } for endpoint ${endpoint} successful quote: ${request.amount.toString()} -> ${quote.toString()}}`
-      );
+      });
```
rileydcampbell, line 199: Do these firehouse logs show up in CloudWatch? If not, let's keep these info logs because they have been super useful for debugging issues.

lib/quoters/WebhookQuoter.ts
```diff
@@ -200,13 +216,32 @@ export class WebhookQuoter implements Quoter {
...
       metric.putMetric(metricContext(Metric.RFQ_FAIL_ERROR, name), 1, MetricLoggerUnit.Count);
+      const errorLatency = {
+        responseTime: Date.now(),
+        latencyMs: Date.now() - before,
+      };
       if (e instanceof AxiosError) {
-        this.log.error(
-          { endpoint, status: e.response?.status?.toString() },
-          `Axios error fetching quote from ${endpoint}: ${e}`
-        );
```
rileydcampbell, line 207: Same note here on keeping for debugging purposes.

lib/quoters/WebhookQuoter.ts
```diff
@@ -200,13 +216,32 @@ export class WebhookQuoter implements Quoter {
...
+          eventProperties: {
+            ...requestContext,
+            status: e.response?.status,
+            data: e.response?.data,
+            ...errorLatency,
+            responseType: axiosResponseType,
+          },
+        });
       } else {
-        this.log.error({ endpoint }, `Error fetching quote from ${endpoint}: ${e}`);
```
rileydcampbell, line 209: ditto

lib/repositories/firehose-repository.ts
```diff
@@ -0,0 +1,53 @@
...
+    try {
+      await this.firehose.putRecord(params).promise();
+      return { statusCode: 200, body: 'Record put successful' };
+    } catch (error) {
+      const errorResponse = `Error putting record. ${error}`
+      this.log.error({ error, streamName: this.streamName }, errorResponse)
+      return { statusCode: 500, body: errorResponse };
+    }
+  }
+}
```
rileydcampbell, line 53:
```suggestion
}

```
nit: newline

lib/repositories/firehose-repository.ts
```diff
@@ -0,0 +1,53 @@
...
+      eventType: AnalyticsEventType;
+      eventProperties: { [key: string]: any };
+    }
+  ): Promise<{ statusCode: number; body: string }> {
+    const { eventType, eventProperties } = eventData;
+    const analyticsEvent = new AnalyticsEvent(eventType, eventProperties);
+
+    if (this.isMock) {
+      return { statusCode: 200, body: 'Record put successful in mock' };
+    }
```
rileydcampbell, line 34: nit: would it be possible to remove this and just mock the return value of `this.firehose.putRecord` when we are testing this? if that makes it too tricky, can ignore this comment

lib/repositories/firehose-repository.ts
```diff
@@ -0,0 +1,53 @@
+import { Firehose } from 'aws-sdk';
+import { AnalyticsEvent, AnalyticsEventType } from '../entities/analytics-events';
+import { default as Logger } from 'bunyan';
+
+export class FirehoseLogger {
```
rileydcampbell, line 5: Might be nice to make this implement an interface like `IAnalyticsLogger` or something. Then we have a nice interface if we ever swapped from using firehose. We have a similar pattern [here](https://github.com/Uniswap/uniswapx-parameterization-api/blob/88db0607dea1ff620db4ae612ac819f1369a74de/lib/providers/webhook/s3.ts#L10).

lib/entities/analytics-events.ts
```diff
@@ -0,0 +1,25 @@
...
+  eventType: AnalyticsEventType;
+  eventTime?: number; // gets set in constructor
+  eventProperties: { [key: string]: any };
+
+  constructor(eventType: AnalyticsEventType, eventProperties: { [key: string]: any }) {
+    this.eventType = eventType;
+    this.eventTime = Date.now();
+    this.eventProperties = eventProperties;
+  }
+};
```
rileydcampbell, line 25:
```suggestion
};

```
nit: newline

lib/quoters/WebhookQuoter.ts
```diff
@@ -119,68 +132,71 @@ export class WebhookQuoter implements Quoter {
...
 
       // RFQ provider explicitly elected not to quote
       if (isNonQuote(request, hookResponse, response)) {
         metric.putMetric(Metric.RFQ_NON_QUOTE, 1, MetricLoggerUnit.Count);
         metric.putMetric(metricContext(Metric.RFQ_NON_QUOTE, name), 1, MetricLoggerUnit.Count);
-        this.log.info(
-          {
-            response: hookResponse.data,
-            responseStatus: hookResponse.status,
+        await this.firehose.sendAnalyticsEvent({
```
rileydcampbell, line 148: One question here, do you have any suspicions on how this will effect overall latency? With our old logs we didn't have to `await` so they didn't effect our overall endpoint response time but looks like these might change that a little bit?

rileydcampbell (commented):
overall lgtm! thanks Blair! 

My main notes are in the Webhook provider. We've deleted some of the `log.info` and `log.error` in favor of using the firehose logger. I think we might want to keep these if the firehose event logging doesn't show up nicely in CloudWatch logs bc the info and error logs were mainly for debugging.

lib/providers/analytics/firehose.ts
rileydcampbell, line 1: Not gonna make this blocking but would be nice if we had a super small unit test `firehose.test.ts` for this file.

rileydcampbell (approved):
lgtm!

PR #246, opened 2024-01-03
feat: Automated CB based on timestamps, Cron and DB part

ConjunctiveNormalForm:
creates a `Timestamp` Dynamo table that tracks filler hash, lastCheckedPostTimestamp (last time the row is updated), and the timestamp that each filler should be blocked until.

The fade-rate cron job now calculates related timestamps based on recent order data from Redshift and `Timestamp` table

cb

### DIFF
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -12,7 +12,7 @@ export const BETA_COMPLIANCE_S3_KEY = 'beta.json';
 export const DYNAMO_TABLE_NAME = {
   FADES: 'Fades',
   SYNTHETIC_SWITCH_TABLE: 'SyntheticSwitchTable',
-  TIMESTAMP: 'Timestamp',
+  FILLER_CB_TIMESTAMPS: 'FillerCBTimestamps',
 };
 
 export const DYNAMO_TABLE_KEY = {
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -24,5 +24,7 @@ export const DYNAMO_TABLE_KEY = {
   TRADE_TYPE: 'type',
   LOWER: 'lower',
   ENABLED: 'enabled',
-  TIMESTAMP: 'timestamp',
+  BLOCK_UNTIL_TIMESTAMP: 'blockUntilTimestamp',
+  LAST_POST_TIMESTAMP: 'lastPostTimestamp',
+  FADED: 'faded',
 };
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -12,6 +12,7 @@ export const BETA_COMPLIANCE_S3_KEY = 'beta.json';
 export const DYNAMO_TABLE_NAME = {
   FADES: 'Fades',
   SYNTHETIC_SWITCH_TABLE: 'SyntheticSwitchTable',
+  TIMESTAMP: 'Timestamp',
 };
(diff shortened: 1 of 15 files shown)


lib/constants.ts
```diff
@@ -12,6 +12,7 @@ export const BETA_COMPLIANCE_S3_KEY = 'beta.json';
 export const DYNAMO_TABLE_NAME = {
   FADES: 'Fades',
   SYNTHETIC_SWITCH_TABLE: 'SyntheticSwitchTable',
+  TIMESTAMP: 'Timestamp',
```
zhongeric, line 15: nit but at first glance this table name wasn't very descriptive for me

test/crons/fade-rate.test.ts
```diff
@@ -1,33 +1,103 @@
...
 
 describe('FadeRateCron test', () => {
-  describe('calculateFillerFadeRates', () => {
+  let newFades: FillerFades;
+  beforeAll(() => {
+    newFades = getFillersNewFades(FADES_ROWS, ADDRESS_TO_FILLER, FILLER_TIMESTAMPS, logger);
+  });
+
+  describe('getFillersNewFades', () => {
     it('takes into account multiple filler addresses of the same filler', () => {
```
zhongeric, line 62: this test name seems outdated?

lib/cron/fade-rate.ts
```diff
@@ -40,49 +51,92 @@ async function main(metrics: MetricsLogger) {
...
     const addressToFillerHash = await webhookProvider.addressToFillerHash();
-    const fillerFadeRate = calculateFillerFadeRates(result, addressToFillerHash, log);
-    log.info({ fadeRates: [...fillerFadeRate.entries()] }, 'filler fade rate');
 
-    const configProvider = new S3CircuitBreakerConfigurationProvider(
-      log,
-      `${FADE_RATE_BUCKET}-prod-1`,
-      FADE_RATE_S3_KEY
+    // aggregated # of fades by filler entity (not address)
+    //  | hash    |  faded  |
```
zhongeric, line 69: nit: fades

zhongeric (approved):
Looks great!

rileydcampbell (approved):
nice nice, this lgtm! just for a thought experiment, if we wanted to create a reset button it could just write to the filler row and set the `blockUntilTimestamp` to nothing right?

ConjunctiveNormalForm:
> nice nice, this lgtm! just for a thought experiment, if we wanted to create a reset button it could just write to the filler row and set the `blockUntilTimestamp` to nothing right?

that's right, although I think the main goal of the design is to fully automate things

lib/repositories/fades-repository.ts
```diff
@@ -56,14 +58,15 @@ export class FadesRepository extends BaseRedshiftRepository {
 }
 
 const CREATE_VIEW_SQL = `
-CREATE OR REPLACE VIEW rfqOrdersTimestamp 
+CREATE OR REPLACE VIEW latestRfqs
```
codyborn, line 61: This gets the latest 20 RFQs for each filler right? The limit of 1k would prevent us from going over 50 fillers. Is this 50 filler limit expected for the foreseeable future?

PR #255, opened 2024-01-23
feat(V2): hard quote skeleton + some refactoring around quote handlers and entities

ConjunctiveNormalForm:
Added basic hard quote lambda handler code + schemas etc, although the the bulk of the work is the refactoring of the various entities/schemas/wrapper classes related to Quote:

- `/handlers/quote-v2`: two sub handlers, `indicative` and `hard`
  - `indicative`: mostly the same code from the v1 quote endpoint
  - `hard`: skeleton code at this point

- types:
  - `IndicativeQuoteRequestBody`: Request from URA for indicative auction
  - `V2QuoteResponseBody`: Response body back to URA for indicative auction
  - `HardQuoteRequestBody`: Request from URA for secondary auction, with signed order from user and signature
    - importantly, there is no `...ResponseBody` counterpart
   
  - `V2QuoteRequest`: unified v2 QuoteRequest type wrapping around both the `Indicative...` and `HardQuoteRequestBody`. Note it parses the HardQuote's encodedOrder to get the same interface as the indicative quote

Not reusing the current quote handler because I expect there to be a short period of time where v1 and v2 would co-exist to allow for an easier transition?

Also many linter related changes from `yarn fix`


lib/handlers/quote-v2/schema.ts
```diff
@@ -0,0 +1,57 @@
...
+/* v2 rfq response from filler */
+export const V2RfqResponseJoi = Joi.object({
+  tokenInChainId: FieldValidator.chainId.required(),
+  tokenOutCHainId: FieldValidator.chainId.required(),
+  requestId: FieldValidator.uuid.required(),
+  tokenIn: Joi.string().required(),
+  amountIn: FieldValidator.amount.required(),
+  tokenOut: Joi.string().required(),
+  amountOut: FieldValidator.amount.required(),
+  cosigner: FieldValidator.address.required(),
```
marktoda, line 41: filler shouldn't need to return cosigner; I think we can actually keep the filler interface exactly the same as before?

ConjunctiveNormalForm: yeah I thought it could be helpful if we require them to explicit ack the labs address? but def not required

marktoda (approved):

ConjunctiveNormalForm:
decided to follow @marktoda 's suggestion of keeping the filler interface exactly the same - note for response to URA, we did change from having `chainId` to `tokenIn/OutChainId` fields

v1 behavior was that URA requests includes `tokenIn/tokenOutChainId` fields but response only has a single `chainId` 🤷

PR #260, opened 2024-01-30
fix: add more MM response validation

rileydcampbell:


lib/entities/QuoteResponse.ts
```diff
@@ -43,10 +47,27 @@ export class QuoteResponse implements QuoteResponseData {
   }
 
   public static fromRFQ(request: QuoteRequestData, data: RfqResponse, type: TradeType): ValidatedResponse {
+    let validationError: ValidationError | undefined;
+
     const responseValidation = RfqResponseJoi.validate(data, {
```
ConjunctiveNormalForm, line 52: would be ideal to do the token addr check in here. Maybe a variant of https://github.com/Uniswap/uniswapx-parameterization-api/blob/4ba4e10dab3f170c06be2ddfa584874bb5e8b8ce/lib/handlers/quote/schema.ts#L8 ?

ConjunctiveNormalForm (approved):
please add a todo to improve how validation is done!

PR #262, opened 2024-02-01
[DO-NOT-MERGE] compare

ConjunctiveNormalForm:


marktoda:
I think we just merge this into main, it shouldn't affect the existing service as it's purely on new endpoints

PR #253, opened 2024-01-12
feat: update firehose client to use aws v3 sdk

ConjunctiveNormalForm:
Completely removes `aws-sdk` package which should reduce bundle size and thus lambda cold start time.

PR #268, opened 2024-02-12
[fix] fix firehose logger output by removing base64 encoding

blairmason:
Previously @ConjunctiveNormalForm put out a PR (https://github.com/Uniswap/uniswapx-parameterization-api/pull/253) to update the firehose client used to log quoter endpoint metrics to use aws v3 sdk. In doing so, he added encoding logic on the data being logged that resulted in it not being in plain text new-line JSON in the destination S3 buckets. Removing the `base64` the encoding parameter produces the right format.


marktoda (approved):
nice ty

PR #270, opened 2024-02-22
add hard quote

marktoda:
- feat: add hard quote initial
- fix: tests

### DIFF
--- a/lib/entities/HardQuoteRequest.ts
+++ b/lib/entities/HardQuoteRequest.ts
@@ -1,11 +1,12 @@
 import { TradeType } from '@uniswap/sdk-core';
-import { V2DutchOrder } from '@uniswap/uniswapx-sdk';
+import { UnsignedV2DutchOrder } from '@uniswap/uniswapx-sdk';
 import { BigNumber, utils } from 'ethers';
-import { QuoteRequestDataJSON } from '.';
+
 import { HardQuoteRequestBody } from '../handlers/hard-quote';
+import { QuoteRequestDataJSON } from '.';
 
 export class HardQuoteRequest {
-  public order: V2DutchOrder;
+  public order: UnsignedV2DutchOrder;
 
   public static fromHardRequestBody(_body: HardQuoteRequestBody): HardQuoteRequest {
     // TODO: parse hard request into the same V2 request object format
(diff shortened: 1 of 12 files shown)


lib/entities/HardQuoteRequest.ts
```diff
@@ -0,0 +1,113 @@
...
+      requestId: this.requestId,
+      swapper: utils.getAddress(this.swapper),
+      // switch tokenIn/tokenOut
+      tokenIn: utils.getAddress(this.tokenOut),
+      tokenOut: utils.getAddress(this.tokenIn),
+      amount: this.amount.toString(),
+      // switch tradeType
+      type: TradeType[type],
+      numOutputs: this.numOutputs,
+      ...(this.quoteId && { quoteId: this.quoteId }),
```
zhongeric, line 51: should we change quoteId? Otherwise first quoteId = real quote?

lib/handlers/hard-quote/schema.ts
```diff
@@ -0,0 +1,22 @@
...
+import { FieldValidator } from '../../util/validator';
+
+/* Hard quote request from user */
+export const HardQuoteRequestBodyJoi = Joi.object({
+  requestId: FieldValidator.requestId.required(),
+  quoteId: FieldValidator.uuid.optional(),
+  encodedInnerOrder: Joi.string().required(),
+  innerSig: FieldValidator.rawSignature.required(),
+  tokenInChainId: FieldValidator.chainId.required(),
+  tokenOutChainId: Joi.number().integer().valid(Joi.ref('tokenInChainId')).required(),
```
zhongeric, line 12: nit: why can't we do FieldValidator.chainId here too?

zhongeric (approved):
looks good!

lib/util/validator.ts
```diff
@@ -33,4 +33,13 @@ export class FieldValidator {
   public static readonly tradeType = Joi.string().valid('EXACT_INPUT', 'EXACT_OUTPUT');
 
   public static readonly uuid = Joi.string().guid({ version: 'uuidv4' });
+
+  // A Raw Signature is a common Signature format where the r, s and v
+  //   are concatenated into a 65 byte(130 nibble) DataHexString
+  public static readonly rawSignature = Joi.string().custom((value: string, helpers: CustomHelpers<string>) => {
```
zhongeric, line 39: nit: i thought sig length details were handled under the hood by ethers?

marktoda: they are, but that wouldn't throw a nice early validation result - we do this for several things, put a coarse validation check in joi so it catches obvious errors early

lib/entities/HardQuoteRequest.ts
marktoda: keeping same semantics as the v1 quote, interesting point tho - I think quoters can disintermediate it themselves by storing both quotes under same quoteId and assuming only one or the other will end up as an order

PR #271, opened 2024-02-27
add hard quote logic

marktoda:
This PR adds the internal logic for hard quotes! Effectively:
- Parse the unsigned order passed in
- run a secondary auction round
- validate the auction result
- create cosignerData using auction result and updated timings
- cosign the full order
- post the full order to gouda service
- return encoded full order

Still remaining TODOs:
- use AWS KMS signer instead of a local wallet for cosigner
- integ tests


lib/entities/HardQuoteRequest.ts
```diff
@@ -76,16 +80,24 @@ export class HardQuoteRequest {
     return utils.getAddress(this.order.info.baseOutputs[0].token);
   }
 
+  public get totalOutputAmountStart(): BigNumber {
+    let amount = BigNumber.from(0);
+    for (const output of this.order.info.baseOutputs) {
```
zhongeric, line 85: this assumes all outputs are same token which is not necessarily true right?

lib/handlers/hard-quote/handler.ts
```diff
@@ -57,6 +104,61 @@ export class QuoteHandler extends APIGLambdaHandler<
...
+
+export function getCosignerData(request: HardQuoteRequest, quote: QuoteResponse): CosignerData {
+  const decayStartTime = getDecayStartTime(request.tokenInChainId);
+  // default to open order with the original prices
+  let filler = ethers.constants.AddressZero;
+  let inputAmount = BigNumber.from(0);
+  const outputAmounts = request.order.info.baseOutputs.map(() => BigNumber.from(0));
+
+  // if the quote is better, then increase amounts by the difference
+  if (request.type === TradeType.EXACT_INPUT) {
```
zhongeric, line 119: What's the expected behavior if less outputs are required for EXACT_INPUT but more input is required? Apply the output adjustment but not the input? Wonder if we should explicitly throw / error on this case if its undesired behavior

zhongeric (approved):
looks good! just a few questions

lib/entities/HardQuoteRequest.ts
marktoda: Assumed true anyways for now in main rfq flownsince quoters can only be for one token

lib/handlers/hard-quote/handler.ts
marktoda: Hmm interesting, I think the api contract should disallow changing the EXACT_X side in any case yea

PR #283, opened 2024-03-08
feat: separate hard quote exports

marktoda:
previously hard quote injector was built in the same index exports file
as the indicative quote injector, causing even indicative quote setup to
require a KMS_KEY_ID environment variable. This commit separates the
export into a separate file to decouple the env pulling.

Future improvement should do this same separation for rest of the
lambdas as well instead of having a global handlers/index

### DIFF
--- a/lib/handlers/hard-quote/exports.ts
+++ b/lib/handlers/hard-quote/exports.ts
@@ -1,0 +1,9 @@
+import { QuoteHandler } from './handler';
+import { QuoteInjector } from './injector';
+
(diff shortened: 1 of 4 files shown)


ConjunctiveNormalForm (approved):

lib/handlers/hard-quote/exports.ts
```diff
@@ -0,0 +1,9 @@
+import { QuoteHandler } from './handler';
```
marktoda, line 1: had to move out of `index.ts` else the injector/handler would be built from unit tests, requiring env variables etc. whenever we do `import { QuoteHandler } from 'hard-quote';

PR #284, opened 2024-03-11
fix: decouple handler exports

ConjunctiveNormalForm:
Decouples lambda handlers by moving the exports to individual files, and completes #283


marktoda (approved):

marktoda:
nice ty!

PR #286, opened 2024-03-14
feat: hard quote firehose streams

ConjunctiveNormalForm:
cannot test response logging yet, but the request log did successfully in my local stack


bin/stacks/analytics-stack.ts
```diff
@@ -183,6 +189,18 @@ export class AnalyticsStack extends cdk.NestedStack {
...
 
+    const hardRequestTable = new aws_rs.Table(this, 'HardRequestTable', {
+      cluster: rsCluster,
+      adminUser: creds,
+      databaseName: RS_DATABASE_NAME,
+      tableName: 'HardRequests',
+      tableColumns: [
+        { name: 'requestId', dataType: RS_DATA_TYPES.UUID, distKey: true },
+        { name: 'createdAt', dataType: RS_DATA_TYPES.TIMESTAMP },
+        { name: 'createdAtMs', dataType: RS_DATA_TYPES.TIMESTAMP_MS },
```
zhongeric, line 200: Anything else we want to log here for requests?

lib/handlers/hard-quote/handler.ts
```diff
@@ -47,21 +47,19 @@ export class QuoteHandler extends APIGLambdaHandler<
...
     log.info({
-      eventType: 'HardQuoteRequest',
+      eventType: 'HardRequest',
       body: {
         requestId: request.requestId,
-        tokenInChainId: request.tokenInChainId,
-        tokenOutChainId: request.tokenInChainId,
-        encoded: requestBody.encodedInnerOrder,
-        sig: requestBody.innerSig,
+        quoteId: request.quoteId,
```
zhongeric, line 56: so `quoteId` here matches the indicative `requestId`?

zhongeric (approved):

bin/stacks/analytics-stack.ts
ConjunctiveNormalForm: yeah I was thinking requestId is enough because we can join with the indicative response log, but maybe it's good to log everything just in case

lib/handlers/hard-quote/handler.ts
ConjunctiveNormalForm: not sure if quoteId should exist in hard request tbh... I just saw it's part of the request currently

PR #290, opened 2024-03-25
feat: firehsoe delivery alarms

ConjunctiveNormalForm:
Adding S3 and Redshift delivery alarms for the firehose streams we have.


marktoda (approved):
lgtm nice

PR #308, opened 2024-04-15
fix: longer axios timeout in integ-test

ConjunctiveNormalForm:


lib/providers/order/uniswapxService.ts
```diff
@@ -18,8 +18,9 @@ export class UniswapXServiceProvider implements OrderServiceProvider {
     this.log.info({ orderHash: order.hash() }, 'Posting order to UniswapX Service');
 
     const axiosConfig = {
```
codyborn, line 20: Can you add a comment above line 20 explaining why we need a longer timeout for testing?

codyborn:
I might be missing some context. If this is for axios in the integ-tests, why not update it in the `AxiosUtils` that the integ-tests use?

ConjunctiveNormalForm:
> I might be missing some context. If this is for axios in the integ-tests, why not update it in the `AxiosUtils` that the integ-tests use?

We call param-api in the integ tests, which in turns posts to order service. Updating AxiosUtils only has an effect on the former but not the latter 😢

bin/app.ts
```diff
@@ -243,6 +247,7 @@ envVars['URA_ACCOUNT'] = process.env['URA_ACCOUNT'] || '';
 envVars['BOT_ACCOUNT'] = process.env['BOT_ACCOUNT'] || '';
 envVars['UNISWAP_API'] = process.env['UNISWAP_API'] || '';
 envVars['ORDER_SERVICE_URL'] = process.env['ORDER_SERVICE_URL'] || '';
+envVars['IS_TEST'] = 'true';
```
codyborn, line 250: Why is this hardcoded to true? Why not pull from an env var?

codyborn (approved):
Left some small comments

PR #314, opened 2024-04-18
feat: dynamic filler address repository

ConjunctiveNormalForm:


lib/repositories/filler-address-repository.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  static create(documentClient: DynamoDBDocumentClient): FillerAddressRepository {
+    this.log = Logger.createLogger({
+      name: 'FillerAddressRepository',
+      serializers: Logger.stdSerializers,
+    });
+
+    const addressTable = new Table({
+      name: DYNAMO_TABLE_NAME.FILLER_ADDRESS,
+      partitionKey: 'pk', // generic partition key name to support both filler and address
```
codyborn, line 27: It looks like DynamoDB supports secondary indexes so we don't need to store two separate records just for indexing:
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.html

ConjunctiveNormalForm: Is the suggestion that we store one `filler<>address` relation per item, make one the partition key and the other the primary key of the index? I feel like that's more complex and doesn't bring performance benefits 🤔

lib/quoters/WebhookQuoter.ts
```diff
@@ -241,6 +243,10 @@ export class WebhookQuoter implements Quoter {
       );
 
       //if valid quote, log the opposing side as well
+      // do not await
```
marktoda, line 246: why not? just to not block? does this mean we could get a rejection here and not see it?

lib/quoters/WebhookQuoter.ts
```diff
@@ -241,6 +243,10 @@ export class WebhookQuoter implements Quoter {
       );
 
       //if valid quote, log the opposing side as well
```
marktoda, line 245: this comment should be below

marktoda (approved):

lib/quoters/WebhookQuoter.ts
ConjunctiveNormalForm: yeah just to have as little latency as possible on the hot path. Even if there are occasional rejects I think that's fine?

marktoda (approved):

PR #315, opened 2024-04-19
feat: bubble up order service error

ConjunctiveNormalForm:


lib/providers/order/uniswapxService.ts
```diff
@@ -14,23 +16,41 @@ export class UniswapXServiceProvider implements OrderServiceProvider {
     this.log = _log.child({ quoter: 'UniswapXOrderService' });
   }
 
-  async postOrder(order: Order, signature: string, quoteId?: string): Promise<void> {
+  async postOrder(order: Order, signature: string, quoteId?: string): Promise<ErrorResponse | void> {
```
codyborn, line 19: It feels kinda strange to have a method that only returns errors and nothing on success. Does it make sense to have it return an object too in the case of success and then we can check for error in the caller?

lib/providers/order/uniswapxService.ts
```diff
@@ -14,23 +16,41 @@ export class UniswapXServiceProvider implements OrderServiceProvider {
...
+          orderType: V2_ORDER_TYPE,
+        },
+        axiosConfig
+      );
+      this.log.info({ orderHash: order.hash() }, 'Order posted to UniswapX Service');
+    } catch (e) {
+      if (e instanceof AxiosError) {
+        this.log.error({ error: e.response?.data }, 'Error posting order to UniswapX Service');
+        return {
+          statusCode: e.response?.data.statusCode,
```
codyborn, line 42:
```suggestion
          statusCode: e.response?.data?.statusCode,
```

codyborn: and same below

codyborn (approved):
Left some minor feedback

lib/providers/order/uniswapxService.ts
ConjunctiveNormalForm: `data` isn't optional so should be fine?

PR #319, opened 2024-05-06
feat: add CB provider v2

ConjunctiveNormalForm:
Adding CB provider v2 by merging #246 and #248 to main. #246 contains more detail on how it works. Note that the quote and hard-quote endpoint still uses v1 at this time.

Also moved `getEligibleEndpoints()` to be part of CB provider so that Quoters don't need to worry about the implementation.


lib/providers/circuit-breaker/dynamo.ts
```diff
@@ -0,0 +1,79 @@
...
+  /* add filler if it's not blocked until a future timestamp */
+  async getEligibleEndpoints(endpoints: WebhookConfiguration[]): Promise<WebhookConfiguration[]> {
+    try {
+      const now = Math.floor(Date.now() / 1000);
+      const fillerTimestamps = await this.getConfigurations();
+      if (fillerTimestamps.size) {
+        this.log.info({ fillerTimestamps: [...fillerTimestamps.entries()] }, `Circuit breaker config used`);
+        const enabledEndpoints: WebhookConfiguration[] = [];
+        endpoints.forEach((e) => {
+          if (
```
marktoda, line 60: nit: maybe more readable to do a filter?

```
endpoints.filter((e) => 
return !(fillerTimestamps.has(e.hash) && blockUntilTimestamp > now);
```

lib/providers/circuit-breaker/mock.ts
```diff
@@ -1,9 +1,59 @@
...
+  }
 
   async getConfigurations(): Promise<CircuitBreakerConfiguration[]> {
     return this.config;
   }
+
+  async getEligibleEndpoints(endpoints: WebhookConfiguration[]): Promise<WebhookConfiguration[]> {
+    const fillerToConfigMap = new Map(this.config.map((c) => [c.hash, c]));
+    const enabledEndpoints: WebhookConfiguration[] = [];
+    endpoints.forEach((e) => {
```
marktoda, line 48: same on filter

lib/providers/circuit-breaker/dynamo.ts
```diff
@@ -0,0 +1,79 @@
...
+import { BaseTimestampRepository, FillerTimestampMap, TimestampRepository } from '../../repositories';
+import { WebhookConfiguration } from '../webhook';
+
+export class DynamoCircuitBreakerConfigurationProvider implements CircuitBreakerConfigurationProvider {
+  private log: Logger;
+  private fillerHashes: string[];
+  private lastUpdatedTimestamp: number;
+  private timestampDB: BaseTimestampRepository;
+  private timestamps: FillerTimestampMap = new Map();
+  allow_list: Set<string> = new Set<string>();
```
marktoda, line 15: not using allow list here at all? also allow_list not super clear to me what its purpose is - comment would be helpful

marktoda (approved):
Looks awesome to me

PR #326, opened 2024-05-17
feat: Use dynamic filler address mapping in fade-rate-v2 cron

ConjunctiveNormalForm:
With #314 we started dynamically tracking addresses of each filler, but we've yet to use them. This PR replaces the static filler -> address config we store in S3 with the dynamic mapping we've been tracking in dynamoDB.

### DIFF
--- a/lib/repositories/fades-repository.ts
+++ b/lib/repositories/fades-repository.ts
@@ -99,8 +99,8 @@ export class V2FadesRepository extends BaseRedshiftRepository {
     const formattedResult = result.map((row) => {
       const formattedRow: V2FadesRowType = {
         fillerAddress: row[0].stringValue as string,
-        faded: Number(row[1].longValue as number),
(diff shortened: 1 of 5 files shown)


lib/repositories/fades-repository.ts
```diff
@@ -99,8 +99,8 @@ export class V2FadesRepository extends BaseRedshiftRepository {
     const formattedResult = result.map((row) => {
       const formattedRow: V2FadesRowType = {
         fillerAddress: row[0].stringValue as string,
-        faded: Number(row[1].longValue as number),
-        postTimestamp: Number(row[2].longValue as number),
```
ConjunctiveNormalForm, line 103: this ordering has to match the one in the SQL query

codyborn: Can you add this as a comment?

codyborn (approved):
Looks great!

PR #327, opened 2024-05-21
Add Arbitrum to ChainId validation

codyborn:


lib/util/chains.ts
```diff
@@ -11,6 +12,7 @@ export enum ChainName {
   GÖRLI = 'goerli',
   POLYGON = 'polygon',
   SEPOLIA = 'sepolia',
+  ARBITRUM_ONE = 'arbitrum-mainnet',
```
ConjunctiveNormalForm, line 15: nit: should value string matches the key?

ConjunctiveNormalForm (approved):

PR #330, opened 2024-05-30
feat: add fillerName and endpoint to quote response logs

ConjunctiveNormalForm:
completing [PROTO-248](https://linear.app/uniswap/issue/PROTO-248/dynamic-filler-mapping-in-dashboard)


codyborn (approved):
Looks great! 🚢

PR #331, opened 2024-05-31
Add Open Order QuoteResponse log

codyborn:
Add QuoteResponse logs for Open Orders

https://linear.app/uniswap/issue/PROTO-256/add-open-order-quoteresponse-log

### DIFF
--- a/lib/handlers/hard-quote/handler.ts
+++ b/lib/handlers/hard-quote/handler.ts
@@ -51,24 +51,23 @@ export class QuoteHandler extends APIGLambdaHandler<
 
     // Instead of decoding the order, we rely on frontend passing in the requestId
     //   from indicative quote
-    const logBody = {
-      requestId: request.requestId,
-      quoteId: request.quoteId,
-      tokenInChainId: request.tokenInChainId,
-      tokenOutChainId: request.tokenOutChainId,
-      tokenIn: request.tokenIn,
-      tokenOut: request.tokenOut,
-      offerer: request.swapper,
-      amount: request.amount.toString(),
-      type: TradeType[request.type],
-      numOutputs: request.numOutputs,
-      cosigner: request.order.info.cosigner,
-      createdAt: timestampInMstoSeconds(start),
-      createdAtMs: start.toString(),
-    }
     log.info({
       eventType: 'HardRequest',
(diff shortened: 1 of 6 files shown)


lib/handlers/hard-quote/handler.ts
```diff
@@ -87,6 +88,19 @@ export class QuoteHandler extends APIGLambdaHandler<
...
       log.info({ cosignerData: cosignerData }, 'open order with default cosignerData');
+
+      // The RFQ responses are logged in getBestQuote() 
+      // we log the Open Orders here
+      log.info({
+        eventType: 'QuoteResponse',
+        body: { 
+          ...logBody,
+          amountIn: request.totalInputAmountStart.toString(),
+          amountOut: request.totalOutputAmountStart.toString(),
```
codyborn, line 99: @ConjunctiveNormalForm can you double check these should be the `amountIn` and `amountOut` values that we log for Open Orders?

lib/handlers/hard-quote/handler.ts
```diff
@@ -51,25 +51,26 @@ export class QuoteHandler extends APIGLambdaHandler<
...
+      amount: request.amount.toString(),
+      type: TradeType[request.type],
+      numOutputs: request.numOutputs,
+      cosigner: request.order.info.cosigner,
+      createdAt: timestampInMstoSeconds(start),
+      createdAtMs: start.toString(),
+    }
     log.info({
       eventType: 'HardRequest',
-      body: {
```
ConjunctiveNormalForm, line 56: What's the reason for moving this outside?

lib/handlers/hard-quote/handler.ts
ConjunctiveNormalForm: maybe better to use the response fields?

codyborn: Great call out. I didn't see that.

lib/handlers/hard-quote/handler.ts
codyborn: It was to use the same object for logging the response. But no longer relevant with the latest changes.

ConjunctiveNormalForm (approved):
lgtm!

PR #333, opened 2024-06-05
fix(cbv2): checksum address

ConjunctiveNormalForm:


lib/cron/fade-rate-v2.ts
```diff
@@ -67,7 +68,10 @@ async function main(metrics: MetricsLogger) {
   if (result) {
     const fillerEndpoints = webhookProvider.fillerEndpoints();
     const addressToFillerMap = await fillerAddressRepo.getAddressToFillerMap(fillerEndpoints);
-    log.info({ addressToFillerMap }, 'address to filler map from dynamo');
+    log.info(
```
ConjunctiveNormalForm, line 71: log won't log Map properly

codyborn (approved):

PR #343, opened 2024-06-13
Log all hard-quote responses as `HardResponse`

codyborn:
- Log all hard-quote responses as HardResponse
- Open Order quoteId should remain as `request.quoteId`


blairmason (approved):
LGTM

PR #356, opened 2024-07-22
Tighten alarms, new dashboards

codyborn:
## Alarms
- 5xx alarms
  - use sum instead of avg
  - set threshold to have caught prod issues
- quotePostErrorAlarmSev2
  - uncomment alarm

## Dashboard
There was no chart in our dashboard that indicated a failure; however the lambda specific charts *did* show the error. Because the lambdas are named dynamically (ex `prod-us-east-2-GoudaParameteriza-HardQuote29A66D69-y7GaCrCf5djc`), I'm unable to reference it statically. Instead, I iterate through all children of the app and create a dashboard for each lambda function like so:

```
scope.node.children
    .filter((service) => service instanceof lambda.Function)
    .map((service) => { 
     ...
```

Due to some AWS errors, I'm unable to test this in my personal AWS account.

### DIFF
--- a/bin/stacks/api-stack.ts
+++ b/bin/stacks/api-stack.ts
@@ -443,7 +443,7 @@ export class APIStack extends cdk.Stack {
         // For this metric 'avg' represents error rate.
         statistic: 'avg',
       }),
-      threshold: 0.05,
+      threshold: 0.03,
       // Beta has much less traffic so is more susceptible to transient errors.
       evaluationPeriods: stage == STAGE.BETA ? 5 : 3,
     });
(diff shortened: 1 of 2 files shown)


bin/stacks/api-stack.ts
```diff
@@ -453,9 +453,9 @@ export class APIStack extends cdk.Stack {
       metric: api.metricServerError({
         period: Duration.minutes(5),
         // For this metric 'avg' represents error rate.
-        statistic: 'avg',
+        statistic: 'sum',
```
zhongeric, line 456: what does sum mean in this context?

zhongeric (approved):

bin/stacks/api-stack.ts
codyborn: It takes the absolute count over the time period, rather than the avg. It makes the spikes more pronounced and easier to alert on. There shouldn't be any reason we have 500 errors so even small amounts are worth looking into.

ConjunctiveNormalForm: makes sense!

ConjunctiveNormalForm (approved):

PR #360, opened 2024-07-26
Fix S3 clock skew

codyborn:
I believe the clock skew problem is caused by a large time delta between when the client is created and when it's used. This change recreates the client on each request.


ConjunctiveNormalForm (approved):
very surprising that the S3Client would just use a stale timestamp tbh, but nice find!

PR #361, opened 2024-07-29
fix: tweak dynamo tables' provisioned capacity

ConjunctiveNormalForm:
fillerAddress table been getting throttled because of overcapacity usage
Screenshot 2024-07-29 at 2 08 22 PM

on the other hand, read capacity usage of fillerCBTimestamps has been consistently under 1


zhongeric (approved):

bin/config.ts
```diff
@@ -3,8 +3,8 @@ import { BillingMode } from 'aws-cdk-lib/aws-dynamodb';
 import { TableCapacityConfig } from './stacks/cron-stack';
 
 export const PROD_TABLE_CAPACITY: TableCapacityConfig = {
-  fillerAddress: { billingMode: BillingMode.PROVISIONED, readCapacity: 20, writeCapacity: 100 },
+  fillerAddress: { billingMode: BillingMode.PROVISIONED, readCapacity: 50, writeCapacity: 200 },
```
codyborn, line 6: Should we bump it up more? In the chart you shared, it does cross these new thresholds too

ConjunctiveNormalForm: hmm yeah I guess it doesn't really matter if we over-provision for a bit

PR #365, opened 2024-08-06
feat: notify blocked fillers of circuit breaker status

ConjunctiveNormalForm:
For every quote request, send blocked fillers their `blockUntilTimestmp`s

Also removing some dead code related v1 circuit breaker


lib/providers/circuit-breaker/dynamo.ts
```diff
@@ -57,20 +58,36 @@ export class DynamoCircuitBreakerConfigurationProvider implements CircuitBreaker
...
 
-      return endpoints;
+      return {
+        enabled: endpoints,
+        disabled: [],
+      };
     } catch (e) {
       this.log.error({ error: e }, `Error getting eligible endpoints, default to returning all`);
-      return endpoints;
+      return {
```
codyborn, line 87: nit: this could be put in a finally and remove the duplication. We can have the `disabledEndpoints` initialized to `[]` outside the try{

codyborn (approved):
🚢

zhongeric (approved):

PR #371, opened 2024-08-27
fix: use allSettled instead so that we dont throw

ConjunctiveNormalForm:


lib/quoters/WebhookQuoter.ts
```diff
@@ -60,7 +61,7 @@ export class WebhookQuoter implements Quoter {
     const quotes = await Promise.all(enabledEndpoints.map((e) => this.fetchQuote(e, request)));
```
codyborn, line 61: Should we use `allSettled` here as well?

lib/quoters/WebhookQuoter.ts
```diff
@@ -60,7 +61,7 @@ export class WebhookQuoter implements Quoter {
...
 
     // should not await and block
-    Promise.all(disabledEndpoints.map((e) => this.notifyBlock(e)));
+    Promise.allSettled(disabledEndpoints.map((e) => this.notifyBlock(e)));
```
codyborn, line 64: Now that you're catching the error in the `notifyBlock` calls, this wouldn't change the behavior, would it?

lib/quoters/WebhookQuoter.ts
zhongeric: i dont think its necessary because fetchQuote can never reject, though good practice to

lib/quoters/WebhookQuoter.ts
zhongeric: yea wouldn't change but I think we should try to do allSettled everywhere possible

zhongeric: for example, here: https://github.com/Uniswap/uniswapx-parameterization-api/pull/371/files#diff-7a29254ba6f168fefbeb5305206da26cda2f763b9d76a415c32f17a6d5498353R124

the second request will not be made if the first axios call fails

zhongeric (approved):

PR #375, opened 2024-09-05
fix: Fix latency issue on GPA for no quotes found

zi-yang-uni:
- Fixes API-486
- Add handler duration metric 
- Update aws-cdk/aws-redshift-alpha to fix local deploy issues (previous version using nodejs 14 which is deprecated)
- We are taking a latency hit (~650ms) on logging error for CustomError.  If CustomError let's serialize it properly by calling `toJSON` method.  Hopefully this fixes the latency issues we are seeing for NoQuote Error
```

Field | Value
-- | --
1728437195260
637423166136:/aws/lambda/GoudaParameterizationStack-QuoteE2906A56-yPW9v679o3QJ
2024/10/09/[8]c3216dd7f073472287ade8193f5bb3cc
2024-10-09T01:26:32.697Z	5950ac19-b888-47d8-9808-90ce56eb7aee	INFO	Error logging took 640ms
5950ac19-b888-47d8-9808-90ce56eb7aee
1728437192697

```

Screenshot 2024-10-08 at 1 55 05 PM

### DIFF
--- a/cdk.context.json
+++ b/cdk.context.json
@@ -50,11 +50,6 @@
     "us-east-2b",
     "us-east-2c"
   ],
-  "availability-zones:account=032111050613:region=us-east-2": [
-    "us-east-2a",
-    "us-east-2b",
-    "us-east-2c"
-  ],
   "availability-zones:account=637423166136:region=us-east-2": [
     "us-east-2a",
     "us-east-2b",
--- a/cdk.context.json
+++ b/cdk.context.json
@@ -54,5 +54,10 @@
     "us-east-2a",
     "us-east-2b",
     "us-east-2c"
+  ],
+  "availability-zones:account=637423166136:region=us-east-2": [
+    "us-east-2a",
+    "us-east-2b",
+    "us-east-2c"
   ]
 }
--- a/cdk.context.json
+++ b/cdk.context.json
@@ -50,7 +50,7 @@
     "us-east-2b",
     "us-east-2c"
   ],
-  "availability-zones:account=637423166136:region=us-east-2": [
+  "availability-zones:account=032111050613:region=us-east-2": [
     "us-east-2a",
     "us-east-2b",
     "us-east-2c"
--- a/lib/handlers/base/api-handler.ts
+++ b/lib/handlers/base/api-handler.ts
@@ -195,10 +195,12 @@ export abstract class APIGLambdaHandler<
               ({ body, statusCode } = handleRequestResult);
             }
           } catch (err) {
-            log.error({ err }, 'Unexpected error in handler');
             if (err instanceof CustomError) {
-              return err.toJSON(id);
+              const errorJson = err.toJSON(id);
+              log.error({ errorJson }, 'Unexpected error in handler');
+              return errorJson;
             }
+            log.error({ err }, 'Unexpected error in handler');
             metric.putMetric(Metric.QUOTE_500, 1, MetricLoggerUnit.Count);
             return INTERNAL_ERROR(id);
           }
--- a/lib/handlers/base/api-handler.ts
+++ b/lib/handlers/base/api-handler.ts
@@ -9,7 +9,7 @@ import {
 import { default as bunyan, default as Logger } from 'bunyan';
 import Joi from 'joi';
 
-import { Metric } from '../../entities';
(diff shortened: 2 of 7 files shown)


ConjunctiveNormalForm:
might be a dumb question - how does this differ from the Lambda duration metric that AWS generates automatically for each lambda?
Screenshot 2024-09-05 at 5 06 37 PM

zi-yang-uni:
@ConjunctiveNormalForm Do we want to cut the latency by the different handlers? I think the lambda duration is total time since start of the lambda execution, while I think we should still be interested in the latency taken by the actual handlers.

lib/handlers/base/api-handler.ts
```diff
@@ -88,22 +88,30 @@ export abstract class APIGLambdaHandler<
...
+        async (event: APIGatewayProxyEvent, context: Context): Promise<APIGatewayProxyResult> => {
+          const requestStart = new Date().getTime();
+          const handler = this.buildHandler();
+          const response = await handler(event, context);
+          const requestEnd = new Date().getTime();
 
-      const response = await handler(event, context);
+          // Track handler duration
+          metric.putDimensions({ [MetricDimension.METHOD]: this.handlerName });
+          metric.putMetric(Metric.HANDLER_DURATION, requestEnd - requestStart, MetricLoggerUnit.Milliseconds);
```
dannydaniil, line 101: Do we need to await this to make sure it gets logged?

dannydaniil (changes requested):
looks good! Comment for awaiting, but also lets post a SS of this working from your local? These dimensions get nasty so we should make sure it works before we merged!

ConjunctiveNormalForm (approved):

lib/handlers/base/api-handler.ts
zi-yang-uni: putMetric doesnt return a promise so no need to await

lib/handlers/base/api-handler.ts
```diff
@@ -187,10 +195,12 @@ export abstract class APIGLambdaHandler<
               ({ body, statusCode } = handleRequestResult);
             }
           } catch (err) {
-            log.error({ err }, 'Unexpected error in handler');
```
zi-yang-uni, line 190: seeing serialization issue here if err is `CustomError` with logging the error taking up to 650 ms

lib/handlers/base/api-handler.ts
```diff
@@ -187,10 +195,12 @@ export abstract class APIGLambdaHandler<
...
             if (err instanceof CustomError) {
-              return err.toJSON(id);
+              const errorJson = err.toJSON(id);
+              log.error({ errorJson }, 'Unexpected error in handler');
+              return errorJson;
```
ConjunctiveNormalForm, line 201: is this fix suggesting that `return err.toJSON(id)` somehow is significantly more expensive than `const errJson = err.toJSON(id); return errJson;`?

cgkol (approved):

lib/handlers/base/api-handler.ts
zi-yang-uni: i just tried to maintain same behavior by logging the err.toJSON(id)

ConjunctiveNormalForm (approved):
huge!

codyborn (approved):
🤞

dannydaniil (commented):
looks good, one comment

cdk.context.json
```diff
@@ -54,5 +54,10 @@
     "us-east-2a",
     "us-east-2b",
     "us-east-2c"
+  ],
+  "availability-zones:account=637423166136:region=us-east-2": [
+    "us-east-2a",
+    "us-east-2b",
+    "us-east-2c"
```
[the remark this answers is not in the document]
zi-yang-uni: we dont, removed.  this points to my local account

cdk.context.json
```diff
@@ -50,7 +50,7 @@
     "us-east-2b",
     "us-east-2c"
   ],
-  "availability-zones:account=032111050613:region=us-east-2": [
+  "availability-zones:account=637423166136:region=us-east-2": [
```
dannydaniil, line 53: still different here!

dannydaniil (approved):
lgtm

PR #382, opened 2024-11-01
URA Quote Config log pipeline

codyborn:


bin/stacks/analytics-stack.ts
```diff
@@ -1207,6 +1276,20 @@ export class AnalyticsStack extends cdk.NestedStack {
...
+      uraConfigDestination.destinationPolicy = JSON.stringify({
+        Version: '2012-10-17',
+        Statement: [
+          {
+            Sid: '',
+            Effect: 'Allow',
+            Principal: {
+              AWS: props.envVars['URA_ACCOUNT'],
+            },
+            Action: 'logs:PutSubscriptionFilter',
```
ConjunctiveNormalForm, line 1288: I just realized, one log group (corresponds to one lambda in our setup) can only have 2 subscription filters max. Are we still good here?

ConjunctiveNormalForm (approved):

bin/stacks/analytics-stack.ts
codyborn: Thanks for flagging this Zach! I wasn't aware of this limit. I decided instead to combine the quote config log with the existing response log.
https://github.com/Uniswap/uniswapx-parameterization-api/pull/385

PR #384, opened 2024-11-05
feat: Support DutchV3 Orders

alanhwu:


lib/entities/HardQuoteRequest.ts
```diff
@@ -8,19 +8,26 @@ import { HardQuoteRequestBody } from '../handlers/hard-quote';
...
-  constructor(_data: HardQuoteRequestBody) {
+  constructor(_data: HardQuoteRequestBody, orderType: OrderType) {
     this.data = {
       ..._data,
       requestId: _data.quoteId ?? uuidv4(),
     };
-    this.order = UnsignedV2DutchOrder.parse(_data.encodedInnerOrder, _data.tokenInChainId);
+    if (orderType === OrderType.Dutch_V2) {
+      this.order = UnsignedV2DutchOrder.parse(_data.encodedInnerOrder, _data.tokenInChainId);
+    // } else if (orderType === OrderType.Dutch_V3) {
```
codyborn, line 25: this is intended to be uncommented, right?

lib/entities/HardQuoteRequest.ts
```diff
@@ -110,9 +117,20 @@ export class HardQuoteRequest {
...
   public get type(): TradeType {
-    return this.order.info.input.startAmount.eq(this.order.info.input.endAmount)
-      ? TradeType.EXACT_INPUT
-      : TradeType.EXACT_OUTPUT;
+    if (this.order instanceof UnsignedV2DutchOrder) {
+      return this.order.info.input.startAmount.eq(this.order.info.input.endAmount)
+        ? TradeType.EXACT_INPUT
+        : TradeType.EXACT_OUTPUT
+    } 
+    // else if (this.order instanceof UnsignedV3DutchOrder) {
```
codyborn, line 125: + here

lib/entities/V2HardQuoteResponse.ts
```diff
@@ -0,0 +1,42 @@
...
+          amountOut: this.amountOut.toString(),
+          swapper: this.swapper,
+          filler: this.filler,
+          orderHash: this.order.hash(),
+          createdAt: this.createdAt,
+          createdAtMs: this.createdAtMs,
+        };
+    }
+
+    public get amountOut(): BigNumber {
```
codyborn, line 24: This returns the current amount on the auction curve (via `getDecayedAmount()` in the sdk). Why don't we do something similar for v3?

lib/handlers/hard-quote/handler.ts
```diff
@@ -161,61 +174,70 @@ export class QuoteHandler extends APIGLambdaHandler<
...
+  orderType: OrderType
+): CosignerData | V3CosignerData {
+  switch (orderType) {
+    case OrderType.Dutch_V2: {
+      const decayStartTime = getDecayStartTime(request.tokenInChainId);
+      let filler = ethers.constants.AddressZero;
+      let inputOverride = BigNumber.from(0);
+      const outputOverrides = request.order.info.outputs.map(() => BigNumber.from(0));
 
-  // if the quote is better, then increase amounts by the difference
```
codyborn, line 171: did you intend to remove these comments?

lib/handlers/hard-quote/handler.ts
```diff
@@ -161,61 +174,70 @@ export class QuoteHandler extends APIGLambdaHandler<
...
+      return v2Data;
     }
-  } else {
-    if (quote.amountIn.lt(request.totalInputAmountStart)) {
-      inputOverride = quote.amountIn;
-      if (quote.filler) {
-        filler = quote.filler;
-      }
+
+    case OrderType.Dutch_V3: {
```
codyborn, line 217: We actually won't be calling `getCosignerData` with Dutch_V3 since we will have the flag `forceOpenOrder` for now. Instead of this, I'd throw an error here if it's v3 since we're not expecting anyone to use v3 with RFQ yet.

lib/handlers/hard-quote/handler.ts
```diff
@@ -240,3 +262,64 @@ function getDecayEndTime(chainId: number, startTime: number): number {
...
+    exclusiveFiller: filler,
+    exclusivityOverrideBps: DEFAULT_EXCLUSIVITY_OVERRIDE_BPS,
+    inputOverride: inputOverride,
+    outputOverrides: outputOverrides,
+  };
+}
+
+function getDefaultV3CosignerData(request: HardQuoteRequest): V3CosignerData {
+  return {
+    decayStartBlock: 1,
```
codyborn, line 319: This is the absolute block number (not relative)

lib/entities/V2HardQuoteResponse.ts
alanhwu: The only place this is used is in logging, and for V3 I just directly logged the input, which gives starting amount and curve

lib/handlers/hard-quote/handler.ts
alanhwu: I'll reintroduce them!

lib/handlers/hard-quote/handler.ts
alanhwu: Did some thinking on this, I think we'll need to add something to injector that we can use the grab the L2 block and then fill this field with `L2Block + V3_BLOCK_BUFFER`

alanhwu: Addressed in latest commit

lib/entities/HardQuoteRequest.ts
```diff
@@ -122,12 +121,12 @@ export class HardQuoteRequest {
...
     } 
-    // else if (this.order instanceof UnsignedV3DutchOrder) {
-    //   const startAmount = this.order.info.input.startAmount;
-    //   return startAmount.eq(V3DutchOrderBuilder.getMinAmountOut(startAmount, this.order.info.input.curve.relativeAmounts))
-    //     ? TradeType.EXACT_INPUT
-    //     : TradeType.EXACT_OUTPUT
-    // } 
+    else if (this.order instanceof UnsignedV3DutchOrder) {
+      const startAmount = this.order.info.input.startAmount;
+      return startAmount.eq(V3DutchOrderBuilder.getMinAmountOut(startAmount, this.order.info.input.curve.relativeAmounts))
```
codyborn, line 126: `this.order.info.input.curve` could be undefined so we can just add that in the check
`!this.order.info.input.curve || startAmount.eq(V3DutchOrderBuilder.getMinAmountOut(startAmount, this.order.info.input.curve.relativeAmounts))`

lib/handlers/hard-quote/handler.ts
```diff
@@ -109,7 +109,7 @@ export class QuoteHandler extends APIGLambdaHandler<
       cosignerData = getCosignerData(request, bestQuote, orderType);
       log.info({ bestQuote: bestQuote }, 'bestQuote');
     } else {
-      cosignerData = getDefaultCosignerData(request, orderType);
+      cosignerData = await getDefaultCosignerData(request, orderType, provider);
```
codyborn, line 112: I think we should do something similar to URA:
`const provider = chainIdRpcMap.get(requestBody.tokenInChainId);`

I know it's only used for arbitrum today but I worry that some dev in the future will call `provider.getBlockNumber()` expecting it to the be the provider of the current chain and not get any kind of error message.

bin/app.ts
```diff
@@ -123,7 +128,7 @@ export class APIPipeline extends Stack {
       envVars: {
         RFQ_WEBHOOK_CONFIG: rfqWebhookConfig.secretValue.toString(),
         ORDER_SERVICE_URL: urlSecrets.secretValueFromJson('GOUDA_SERVICE_BETA').toString(),
-        ARBITRUM_RPC_URL: urlSecrets.secretValueFromJson('ARBITRUM_RPC_URL').toString(),
+        RPC_42161: rpcUrls.secretValueFromJson('RPC_42161').toString(),
```
codyborn, line 131: nit: Just to make it auto-updating we can do something like we do in URA
```
    const jsonRpcProviders = {} as {[chainKey: string]: string};
    ChainConfigManager.getChainIdsByRoutingType(RoutingType.CLASSIC).forEach(
      (chainId: ChainId) => {
        const mapKey = `RPC_${chainId}`;
        jsonRpcProviders[mapKey] = jsonRpcProvidersSecret
          .secretValueFromJson(mapKey)
          .toString();
      }
    );
```

lib/handlers/hard-quote/injector.ts
```diff
@@ -78,15 +78,23 @@ export class QuoteInjector extends ApiInjector<ContainerInjected, RequestInjecte
...
-    const provider = new ethers.providers.JsonRpcProvider(
-      rpcProviderUrl,
+    const chainIdRpcMap = new Map<ChainId, ethers.providers.StaticJsonRpcProvider>();
+    supportedChains.forEach(
+      chainId => {
+        const rpcUrl = checkDefined(
+          process.env[`RPC_${chainId}`],
+          `RPC_${chainId} is not defined`
+        );
+        const provider = new ethers.providers.StaticJsonRpcProvider(rpcUrl, chainId); // specify chainId to avoid detecctNetwork() call on initialization
```
codyborn: nit: `detectNetwork`

codyborn (approved):
Looks great! Nice work

PR #389, opened 2024-11-27
[TO-BE-REVERTED]: Remove all lambdas

codyborn:
Attempt to fix the beta env by removing all lambdas. Will disable transition to prod.

### DIFF
--- a/bin/stacks/api-stack.ts
+++ b/bin/stacks/api-stack.ts
@@ -2,17 +2,17 @@ import * as cdk from 'aws-cdk-lib';
 import { CfnOutput, Duration } from 'aws-cdk-lib';
 import * as aws_apigateway from 'aws-cdk-lib/aws-apigateway';
 import { MethodLoggingLevel } from 'aws-cdk-lib/aws-apigateway';
-import * as aws_asg from 'aws-cdk-lib/aws-applicationautoscaling';
+// import * as aws_asg from 'aws-cdk-lib/aws-applicationautoscaling';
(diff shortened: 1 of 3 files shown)


bin/stacks/api-stack.ts
```diff
@@ -170,7 +171,7 @@ export class APIStack extends cdk.Stack {
      * Firehose Initialization
      */
 
-    const firehoseStack = new FirehoseStack(this, 'FirehoseStack');
+    // const firehoseStack = new FirehoseStack(this, 'FirehoseStack');
```
ConjunctiveNormalForm, line 174: I'd prefer to only make bare minimum amount of changes

bin/stacks/api-stack.ts
```diff
@@ -242,185 +244,185 @@ export class APIStack extends cdk.Stack {
...
-      lambdaRole: lambdaRole,
-      chatbotSNSArn: chatbotSNSArn,
-      stage: stage,
-    });
-
-    new CronDashboardStack(this, 'CronDashboardStack', {
-      synthSwitchLambdaName: cronStack.synthSwitchCronLambda.functionName,
-      quoteLambdaName: quoteLambda.functionName,
-    });
+    // const analyticsStack = new AnalyticsStack(this, 'AnalyticsStack', {
```
ConjunctiveNormalForm, line 404: maybe instead of commenting out the whole stack, we simply use hard-coded logGroup name (so that we don't have to pass in quote/hardQuoteLambda?

PR #396, opened 2025-02-04
Pull optional dynamic compliance list

codyborn:
Adds a new property `complianceListUrl` to the `FillerComplianceConfiguration`:
```
interface FillerComplianceConfiguration {
  endpoints: string[];
  addresses: string[];
  complianceListUrl?: string;
}
```
Every 5 mins this endpoint is polled to get the latest version of the compliance list and added to the list in the static config.


ConjunctiveNormalForm (approved):
lgtm!

PR #399, opened 2025-02-26
GPA BUIDL Support

codyborn:
Adds support for permissioned tokens
1. Ignore permissioned tokens when calculating circuit breaker (see `fades-repository.ts`)
2. Ignore circuit breaker if trade contains a permissioned token (see `WebhookQuoter.ts`)
3. Ignore quotes from fillers if they're unfillable (see `QuoteResponse.ts`)


lib/quoters/WebhookQuoter.ts
```diff
@@ -39,10 +41,14 @@ export class WebhookQuoter implements Quoter {
...
 
-  public async quote(request: QuoteRequest): Promise<QuoteResponse[]> {
+  public async quote(request: QuoteRequest, provider?: ethers.providers.JsonRpcProvider): Promise<QuoteResponse[]> {
     const statuses = await this.getEndpointStatuses();
     const endpointToAddrsMap = await this.complianceProvider.getEndpointToExcludedAddrsMap();
-    const enabledEndpoints = statuses.enabled.filter(
+    // Ignore endpoint status if token is permissioned
+    const isPermissionedToken = PermissionedTokenValidator.isPermissionedToken(request.tokenIn, request.tokenInChainId) 
+    || PermissionedTokenValidator.isPermissionedToken(request.tokenOut, request.tokenOutChainId);
+    const baseFillerSet = isPermissionedToken ? statuses.enabled.concat(statuses.disabled.map(e => e.webhook)) : statuses.enabled;
```
alanhwu, line 50: Why include the disabled endpoints too?

lib/entities/QuoteResponse.ts
```diff
@@ -112,7 +151,7 @@ export class QuoteResponse implements QuoteResponseData {
         type,
         metadata
       ),
-      ...(validationError && { validationError }),
+      ...(validationErrors.length > 0 && { validationError: {message: validationErrors.join(',\n'), value: data} }),
```
alanhwu, line 154: might be cleaner to return it in a structured way with a `.map` over the errors array

lib/entities/QuoteResponse.ts
```diff
@@ -100,6 +96,49 @@ export class QuoteResponse implements QuoteResponseData {
...
+          validationErrors.push(`preTransferCheck check failed for tokenIn: ${request.tokenIn} from ${request.swapper} to ${data.filler} with amount ${amountIn}`);
+          }
+        }
+      }
+      if (data.filler && PermissionedTokenValidator.isPermissionedToken(request.tokenOut, request.tokenOutChainId)){
+        if (!provider) {
+          validationErrors.push(`provider is required for permissioned token check for tokenOut: ${request.tokenOut}`);
+        } else {
+          // permissioned token is tokenOut
+          const preTransferCheckResult = await PermissionedTokenValidator.preTransferCheck(
```
alanhwu, line 124: If both tokens are permissioned, we'd be running 2 RPC awaits sequentially. Would it be possible to await them together?

lib/quoters/WebhookQuoter.ts
codyborn: Since we'll have a small number of these permissioned token fillers to start, we want to make sure that we always have market makers, even if they're circuit broken normally.

lib/entities/QuoteResponse.ts
codyborn: I'm not sure I follow.. what's your proposed change? Something like this:
`validationErrors.map(error => `[${error}]`).join('\n')`?

lib/entities/QuoteResponse.ts
codyborn: good idea

alanhwu (approved):

lib/entities/QuoteResponse.ts
```diff
@@ -67,31 +69,25 @@ export class QuoteResponse implements QuoteResponseData {
     );
   }
 
-  public static fromRFQ(args: FromRfqArgs): ValidatedResponse {
+  public static async fromRFQ(args: FromRfqArgs, provider?: ethers.providers.JsonRpcProvider, log?: Logger): Promise<ValidatedResponse> {
     const { request, data, type, metadata } = args;
-    let validationError: ValidationError | undefined;
+    let validationErrors: string[] = [];
```
zhongeric, line 74: Why not do "ValidationError[]", I much prefer the loose typing that we had already over arbitrary strings

lib/entities/QuoteResponse.ts
```diff
@@ -100,6 +96,60 @@ export class QuoteResponse implements QuoteResponseData {
       request.type === TradeType.EXACT_INPUT
         ? [request.amount, BigNumber.from(data.amountOut ?? 0)]
         : [BigNumber.from(data.amountIn ?? 0), request.amount];
+
+    // permissioned tokens check
+    try {
```
zhongeric, line 101: maybe consider breaking this out into a smaller function? Bit hard to read right now with all of the branching

zhongeric: could have a function which returns a promise conditionally on the isPermissionedToken check + data.filler being set, etc. and just Promise.all for tokenIn + tokenOut

lib/entities/QuoteResponse.ts
zhongeric: Re my comment above, generating one long string seems very hard to parse out / consume. We should consider extending the response schema potentially to support an array of errors, or returning the first validationError that is set?

lib/repositories/fades-repository.ts
```diff
@@ -193,6 +194,8 @@ SELECT
     postTimestamp,
     CASE WHEN (decayStartTime < fillTimestamp) THEN 1 ELSE 0 END AS faded
 FROM latestRfqsV2
+WHERE tokenIn NOT IN [${PERMISSIONED_TOKENS.map(token => token.address).join(',')}]
+AND tokenOut NOT IN [${PERMISSIONED_TOKENS.map(token => token.address).join(',')}]
```
zhongeric, line 198: careful of case sensitivity here

lib/entities/QuoteResponse.ts
```diff
@@ -67,31 +69,25 @@ export class QuoteResponse implements QuoteResponseData {
     );
   }
 
-  public static fromRFQ(args: FromRfqArgs): ValidatedResponse {
+  public static async fromRFQ(args: FromRfqArgs, provider?: ethers.providers.JsonRpcProvider, log?: Logger): Promise<ValidatedResponse> {
```
zhongeric, line 72: this is a bit of a meta point but would prefer to not have business logic within these entity classes, it may be cleaner to separate out the _creation_ of a quote request/response from whether it is valid or not from dependencies on external services. That would allow us to not have to have an optional provider field here

lib/entities/QuoteResponse.ts
codyborn: The ValidtionErrors would all have the same value for the `value` field. (`data`). The only thing different is the error strings which is why I collected them as an array of strings.

lib/repositories/fades-repository.ts
codyborn: great catch

lib/entities/QuoteResponse.ts
codyborn: LMK if this is what you had in mind:
https://github.com/Uniswap/uniswapx-parameterization-api/pull/399/commits/d0c19ddf2a4d99f7fcfa70e99ac3b4a3c3b7f7bf#diff-7a29254ba6f168fefbeb5305206da26cda2f763b9d76a415c32f17a6d5498353R166

lib/entities/QuoteResponse.ts
zhongeric: Ah I see, hm that's a bit tricky to deal with

lib/quoters/WebhookQuoter.ts
```diff
@@ -148,12 +157,20 @@ export class WebhookQuoter implements Quoter {
         fillerName: config.name,
       };
 
-      const { response, validationError } = QuoteResponse.fromRFQ({
+      const { response, validationError } = await QuoteResponse.fromRFQ({
```
zhongeric, line 160: remove await here

lib/quoters/WebhookQuoter.ts
```diff
@@ -177,12 +194,13 @@ export class WebhookQuoter implements Quoter {
       }
 
       // RFQ provider response failed validation
-      if (validationError) {
+      if (validationError || validatePermissionedTokensError) {
+        const error = validationError || validatePermissionedTokensError;
         metric.putMetric(Metric.RFQ_FAIL_VALIDATION, 1, MetricLoggerUnit.Count);
         metric.putMetric(metricContext(Metric.RFQ_FAIL_VALIDATION, name), 1, MetricLoggerUnit.Count);
```
zhongeric, line 200: nit: should we track permissioned token related errors separately? I don't know if we'd want to alarm on them, for example

lib/util/rfqValidator.ts
```diff
@@ -0,0 +1,104 @@
...
+    to: string,
+    amount: string,
+    provider?: ethers.providers.JsonRpcProvider
+  ): Promise<string | undefined> {
+    if (!PermissionedTokenValidator.isPermissionedToken(tokenAddress, chainId)) {
+      return undefined;
+    }
+
+    if (!provider) {
+      return `provider is required for permissioned token check for token: ${tokenAddress} on chain: ${chainId}`;
```
zhongeric, line 32: if we end up doing custom errors would be nice to add them here to make the typing more strict

lib/quoters/WebhookQuoter.ts
```diff
@@ -247,16 +265,25 @@ export class WebhookQuoter implements Quoter {
...
         metadata,
       });
+      const validateOpposingPermissionedTokensError = await RFQValidator.validatePermissionedTokens(
+        opposingRequest,
+        opposite.data,
+        opposingRequest.amount,
+        opposingResponse.response.amountOut,
+        provider,
+        this.log
+      );
```
zhongeric, line 281: What is the desired behavior for the opposing request, since it's not technically the real quote? If the validation is done on the amount of the trade this might cause the opposing quote to fail validation when the real one is ok?

codyborn: That's a good question. It technically could be the case that the opposing is not permissioned. Given the current state of permissioned tokens, I think it'd make sense to skip sending the opposing quote altogether.

codyborn: Going to keep the request there to not make it too complicated. We just wont perform the permission check on the opposing order.

lib/util/rfqValidator.ts
codyborn: LMK if this is what you had in mind

lib/quoters/WebhookQuoter.ts
codyborn: I don't see any alarms on this event. I think we're good to log them together.

zhongeric (approved):

PR #402, opened 2025-03-06
Add API keys for hard-quote endpoint

codyborn:
Lock down hard-quote to only Trading API. Generate a dev key for Frequent Swapper, UniswapX-tool, and E2E tests.

### DIFF
--- a/bin/stacks/api-stack.ts
+++ b/bin/stacks/api-stack.ts
@@ -366,7 +366,10 @@ export class APIStack extends cdk.Stack {
         allowMethods: aws_apigateway.Cors.ALL_METHODS,
       },
     });
-    hardQuote.addMethod('POST', hardQuoteLambdaIntegration);
+
+    hardQuote.addMethod('POST', hardQuoteLambdaIntegration, { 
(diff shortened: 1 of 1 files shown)


bin/stacks/api-stack.ts
```diff
@@ -377,12 +380,16 @@ export class APIStack extends cdk.Stack {
     });
     const enabled = switchResource.addResource('enabled');
 
-    /* add auth key */
+    /* add auth keys */
+    const tradingAPIKey = api.addApiKey('TradingAPIKey');
+    const devAPIKey = api.addApiKey('DevAPIKey');
     const apiAuthzKey = api.addApiKey('AuthzKey');
```
zhongeric, line 386: I may have missed this but what is authz?

zhongeric (approved):

bin/stacks/api-stack.ts
codyborn: It looks like it was used for the "synthetic switch" lambda, see this line:
`enabled.addMethod('GET', switchLambdaIntegration, { apiKeyRequired: true });`

I don't think the synthetic switch is still being used.

PR #407, opened 2025-04-14
bug(DutchV3): Fix bug in exact_out detection

codyborn:
For EXACT_OUT orders, the input relative amounts are negative. This means that `getMinAmountOut()` will return the start amount if you pass in the input. Instead, to detect EXACT_IN, I check that either the input curve is undefined or all the values are 0.


alanhwu (approved):
just need to fix the lint issue

PR #436, opened 2026-04-06
fix: provide quoteId for open orders and improve postOrder error handling

codyborn:
## Summary
- The `forceOpenOrder` integ test was failing in CI with an opaque 400 response (`{ id: '...' }`, no error detail)
- **Root cause 1**: The `forceOpenOrder` path sent `quoteId: undefined` to the order service. The comment on line 122 says "create random new quoteId" but the code didn't follow through. Now falls back to `requestId` as quoteId. Also added `quoteId` to the integ test request body.
- **Root cause 2**: `postOrder` error handling read `statusCode` from `e.response.data.statusCode` (response body) instead of `e.response.status` (HTTP status). On timeouts or non-standard error formats, all error fields were `undefined`, producing an opaque error with no detail.

## Test plan
- [x] Unit tests pass locally (145/145)
- [x] Integ tests pass locally (11/11)
- [ ] CI integ tests pass in deployment pipeline

### DIFF
--- a/test/integ/hard-quote.test.ts
+++ b/test/integ/hard-quote.test.ts
@@ -175,6 +175,7 @@ describe('Hard Quote endpoint integration test', function () {
 
       const quoteReq: HardQuoteRequestBody = {
         requestId: REQUEST_ID,
+        quoteId: uuidv4(),
         encodedInnerOrder: v2Order.serialize(),
         innerSig: signature,
         tokenInChainId: SEPOLIA,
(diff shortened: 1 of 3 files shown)


test/integ/hard-quote.test.ts
```diff
@@ -164,9 +164,9 @@ describe('Hard Quote endpoint integration test', function () {
       const prebuildOrder = builder
         .input({ token: TOKEN_IN, startAmount: AMOUNT, endAmount: AMOUNT })
         .output({ token: TOKEN_OUT, startAmount: AMOUNT, endAmount: AMOUNT, recipient: SWAPPER_ADDRESS })
-        .nonce(BigNumber.from(Math.floor(Math.random() * 1e15)))
+        .nonce(BigNumber.from(100))
```
ericneil-sanc, line 167: What is the reason behind no more random nonce?

ericneil-sanc (approved):

test/integ/hard-quote.test.ts
codyborn: I added it as a previous fix to these tests but they're still failing

PR #438, opened 2026-05-02
feat: V3 RFQ cosigning + register 14 AMM chains

codyborn:
## Summary

- **Implements DutchV3 RFQ cosigning.** Previously `getCosignerData` for `Dutch_V3` threw `Unsupported order type`; only the *default* V3 path worked. New case mirrors the V2 RFQ flow and handles multi-output orders (swapper-facing output at index 0 absorbs the quote improvement; fee outputs at higher indexes stay at the reactor's sentinel-zero for "use baseOutput").
- Multichain chain registration in `lib/util/chains.ts` (`ChainId` enum, `supportedChains`) and `lib/config/chains.ts` (Joi `chainId` validator). Adds Optimism, BNB, Unichain, Polygon, Monad, XLayer, Worldchain, Soneium, Zora, Base, Celo, Avalanche, Blast. Tempo registered.
- V3 timing derived from wallclock targets, mirroring `getDecayBlockLength`:
  - `V3_DEFAULT_DECAY_DURATION_SECS = 30` + `getDecayBlockLength(chainId) = ceil(30 / blockTimeSecs)`.
  - `V3_DECAY_START_BUFFER_SECS = 1` + `getV3BlockBuffer(chainId) = ceil(1 / blockTimeSecs)`.
- `getBlockTimeSecs(chainId)` populated for all rollout chains; throws on unknown chainIds (rather than silently returning Mainnet's 12s, which would undercount on any new sub-second chain).
- `getWebhookTimeoutMs(chainId)` per-chain timeout: Mainnet 500ms, others 250ms. Used in `WebhookQuoter` axios config.
- `getRpcUrl(chainId)` helper in `lib/util/chains.ts`: reads `RPC_<chainId>` if set (override), else `${RPC_PREFIX_URL}/${chainId}`. Throws if neither.
- `bin/app.ts` wires `RPC_PREFIX_URL` through both the deployed-Lambda env (sourced from `rpcUrls` Secrets Manager) and the local-dev `jsonRpcProviders` map (sourced from `process.env`).
- Bumps `@uniswap/uniswapx-sdk` to `3.0.3` (multi-chain V3 reactor + quoter mappings).
- Dead-code removal: deprecated `WEBHOOK_TIMEOUT_MS` export, `ChainName` enum, `ID_TO_NETWORK_NAME` switch, and `getFillerSupportedProtocols` on the webhook providers (zero callers).

## Implementation notes

- V3 cosigner-amount invariants are the **same direction** as V2 (`inputOverride ≤ baseInput.startAmount`, `outputOverride ≥ baseOutput.startAmount`). Both enforce swapper-improvement; verified against `V3DutchOrderReactor._updateWithCosignerAmounts`. The reactor enforces them per-output on-chain, so the cosigner doesn't re-validate.
- `adjustmentPerGweiBaseFee` and `startingBaseFee` live on the swapper-signed `UnsignedV3DutchOrderInfo`, not `V3CosignerData`. Cosigner can't zero them out at signing time — the lever is on the order-construction side (`DutchV3OrderFactory` in trading-api).
- `getCosignerData` is now `async` and accepts an optional `provider` (required for V3). Internal callers updated.
- `lib/cron/fade-rate-v2.ts` deliberately untouched — circuit-breaker logic is intentionally chain-agnostic.

## Context

Part of Linear [ECO-365](https://linear.app/uniswap/issue/ECO-365). Companion PRs in x-contracts, uniswapx-sdk, x-service, trading-api.

## Test plan

- [x] `yarn lint`: clean
- [x] `yarn test:unit`: 171/171 passed
- [ ] Reviewer: security review of the V3 RFQ cosigning branch (override validation, decay-block math, signature path)
- [ ] Reviewer: per-chain `getBlockTimeSecs` values and the 1s decay-start buffer


codyborn:
## Code Review — PR #438 (V3 RFQ cosigning + Tempo)

**Diff**: `+385/-13` across 7 files; mergeable. This is the substantive PR of the cross-repo set.

### Overview
Implements DutchV3 RFQ cosigning (previously threw `Unsupported order type`), adds Tempo (4217), and unifies V3 decay duration as wallclock-time so per-chain block counts derive from `getBlockTimeSecs(chainId)` × `V3_DEFAULT_DECAY_DURATION_SECS`. Per-chain `V3_BLOCK_BUFFER` map preserves existing chain behavior and slots Tempo at 1.

### Strengths
- The V3 RFQ branch correctly mirrors the V2 swapper-improvement flow (Correction A in the playbook). Inline comments cite the on-chain invariants.
- `getCosignerData` is now `async` with optional `provider`; sensible compromise that keeps V2 callers simple.
- New `handlerDutchV3.cosigner.test.ts` covers both `TradeType` directions, both pass and fail invariant cases, and Tempo-specific paths (buffer=1, baseFee read, null-tolerance).
- `getDecayBlockLength` math is centrally tested (`mainnet=3, arbitrum=120, tempo=60`) — clean and chain-agnostic.
- Backwards compatibility: scalar `V3_BLOCK_BUFFER = 4` export retained; only one internal caller refactored.

### Issues / Suggestions

**Possible correctness bug — multi-output orders:**
- In the EXACT_INPUT branch, `proposedOutput = baseOutputStart + (quote.amountOut - totalOutputAmountStart)` only equals `quote.amountOut` when there is a single output. For multi-output V3 orders, `totalOutputAmountStart` sums all outputs but `baseOutputStart` is just `outputs[0].startAmount`. The arithmetic produces an incorrect override for `outputs[0]` and leaves `outputs[1..n]` at zero (which the reactor will treat as "use base"). Confirm whether multi-output V3 RFQ is in scope; if it is, fix the math. If it isn't, add an explicit guard `if (request.order.info.outputs.length > 1) throw` and a comment.
- Compare to the V2 branch above (`outputs.length === 1` is implicit there too — same potential issue, but separate concern).

**Dead branch:**
- `if (proposedOutput.gte(baseOutputStart))` (line ~265) is always `true` because the parent branch already proved `quote.amountOut > totalOutputAmountStart`, so `increase > 0`, so `baseOutputStart + increase >= baseOutputStart`. Either remove the redundant check or change it to a real invariant assertion (e.g., `<= baseInput.maxAmount` analog). The comment above promises the invariant check, but the code doesn't actually exercise it.

**Code smell — `void _startingBaseFee`:**
- The Tempo branch reads `latest.baseFeePerGas` then explicitly discards it via `void`. Documented as a placeholder for future tripwire/logging, but right now it's a dead read that costs an RPC call per Tempo cosignature. Either:
  1. Remove until needed (replace with `// TODO(TRA2-12): observe baseFee here once tripwire policy is decided`), or
  2. Promote to a `metrics.putMetric('TempoBaseFee', ...)` so the read has observable value.

**Test type safety regression:**
- Existing V2 tests now use `cosignerData: any = await getCosignerData(...)` (lines 319, 333, 346, 360 in the diff). Was `CosignerData`. Type assertion to `as CosignerData` is preferable to `any` — preserves compile-time guarantees on subsequent property access.

**Test setup fragility:**
- `buildV3Order` + `Object.create(HardQuoteRequest.prototype)` bypasses normal SDK construction because uniswapx-sdk doesn't yet publish a Tempo permit2/reactor mapping (per #577). It works, but the production code path is untested at this layer. Add a TODO to simplify once #577 lands and a real `HardQuoteRequest.fromHardRequestBody` works for Tempo.

**Coverage gap:**
- No Tempo test case covers the *bestQuote applied* path (override actually set). The 2 Tempo tests only check `decayStartBlock` and `getBlock('latest')` invocation. Recommend adding a `EXACT_INPUT, better quote on Tempo → outputOverride > 0 and exclusiveFiller set` case symmetric to the Arbitrum one.

**Deprecation hygiene:**
- `export const V3_BLOCK_BUFFER = V3_BLOCK_BUFFER_DEFAULT` is kept "for backwards compatibility." Add a `/** @deprecated use getV3BlockBuffer(chainId) */` JSDoc so external importers (if any) see the deprecation in their IDE.

**Pre-existing tsc errors on main:**
- The PR description notes unrelated tsc failures (`v4PoolSelection`, integ-test globals). If CI gates on `yarn build` rather than `yarn test:unit` directly, those will block this PR's merge. Confirm CI green before approval.

### Project conventions
- Matches existing patterns in `handler.ts` (switch on `OrderType`, mirror V2 structure, helper functions in `lib/constants.ts`).
- `lib/util/chains.ts` enum + `lib/config/chains.ts` `SUPPORTED_CHAINS` both updated — Correction C from the playbook is honored.

### Performance
- One extra `getBlockNumber` + (Tempo only) `getBlock('latest')` per V3 RFQ cosignature. Both reads are fast and cached at the provider; no concerns at expected QPS. The dead Tempo `getBlock` call costs cycles for no benefit until the TODO is resolved (see above).

### Security
- Override invariants enforced at the right layer (cosigner only narrows toward swapper benefit). The reactor enforces the same invariants on-chain, so a buggy cosigner produces a reverting fill, not a stolen one — defense-in-depth holds.
- Verify: a malicious quoter cannot induce the cosigner to set an `exclusiveFiller` that wasn't theirs. Currently, `filler = quote.filler` is taken at face value from the quote; the only gate is "the override is strictly better." If a quoter responds with `quote.filler = otherPMM`, this code happily sets exclusivity to `otherPMM`. Worth confirming this matches V2 trust model — V2 has the same pattern, so probably intentional, but worth an explicit note.

### Verdict
**Approve with revisions.** Fix or guard the multi-output arithmetic (real correctness concern), remove the dead invariant check, address the `void _startingBaseFee` dead read or commit to using it, and tighten test types. The other items are polish.

🤖 Review performed via `/review`

lib/util/chains.ts
```diff
@@ -1,38 +1,58 @@
...
-  ARBITRUM_ONE = 'arbitrum-mainnet',
-}
+/**
+ * Resolve the RPC URL for a given chainId. Per-chain `RPC_<chainId>` env
+ * vars take precedence over the shared `RPC_PREFIX_URL` so individual chains
+ * can be pointed at a different provider when needed; otherwise the chainId
+ * is appended to `RPC_PREFIX_URL` to form the full URL. Throws if neither is
+ * set.
+ */
+export const getRpcUrl = (chainId: number): string => {
```
alanhwu: The hard-quote injector needs to follow this pattern too

lib/constants.ts
```diff
@@ -35,9 +37,69 @@ export const POST_ORDER_ERROR_REASON = {
...
+// Per-chain block time in seconds. Used by getV3BlockBuffer to derive the
+// V3 decay-start buffer in blocks from V3_DECAY_START_BUFFER_SECS.
+export function getBlockTimeSecs(chainId: number): number {
+  switch (chainId) {
+    case 1: // MAINNET
+      return 12;
+    case 10: // OPTIMISM
+      return 2;
+    case 56: // BNB (post-Maxwell hardfork)
+      return 0.5;
```
alanhwu: Should be 0.45

lib/constants.ts
```diff
@@ -35,9 +37,69 @@ export const POST_ORDER_ERROR_REASON = {
...
+    case 130: // UNICHAIN
+      return 1;
+    case 137: // POLYGON
+      return 2;
```
alanhwu: Polygon recently reduced to 1.75

lib/constants.ts
```diff
@@ -35,9 +37,69 @@ export const POST_ORDER_ERROR_REASON = {
...
+    case 143: // MONAD
+      return 0.5;
```
alanhwu: 0.4

lib/constants.ts
```diff
@@ -35,9 +37,69 @@ export const POST_ORDER_ERROR_REASON = {
...
+    case 7777777: // ZORA
+      return 2;
+    case 8453: // BASE
+      return 2;
+    case 42161: // ARBITRUM_ONE
+      return 0.25;
+    case 42220: // CELO
+      return 1;
+    case 43114: // AVALANCHE
+      return 2;
```
alanhwu: It's been about 1s since the last upgrade

lib/handlers/hard-quote/handler.ts
```diff
@@ -258,7 +312,7 @@ function getDecayEndTime(chainId: number, startTime: number): number {
     case ChainId.ARBITRUM_ONE:
       return startTime + 8; // 8 seconds
     default:
-      return startTime + 30; // 30 seconds
+      return startTime + 16; // 16 seconds
```
alanhwu: Calling out that I don't think this change would do anything bc this function is used by Dutch_V2

codyborn: you are correct

lib/util/chains.ts
codyborn: good catch

lib/handlers/hard-quote/handler.ts
```diff
@@ -304,15 +361,35 @@ function getDefaultV2CosignerData(request: HardQuoteRequest): CosignerData {
...
+ * otherwise the resolved output the swapper signed for becomes meaningless —
+ * fills past `decayEndTime` lock in the end-amount, but if `decayEndTime`
+ * is itself past the deadline, the reactor reverts the fill on the
+ * deadline check and the swapper loses the order's exclusivity window
+ * (and the cosigner's exclusivity grant) for nothing.
+ *
+ * V3 orders allow partial decay — fills that happen before
+ * `decayEndBlock` simply interpolate, so this check is V2-only.
+ */
+function assertV2DecayWithinDeadline(decayEndTime: number, deadline: number): void {
```
codyborn, line 384: Separate from main PR: Failing fast with clean error. Today this fails downstream with onchain quote error.

alanhwu (approved):

PR #441, opened 2026-06-04
fix: return 400 instead of retryable 5xx for expired order deadline (EXE-28)

codyborn:
## Summary

Addresses [EXE-28](https://linear.app/uniswap/issue/EXE-28). When a customer submits a `POST /order` whose deadline is too close (or already expired) for the order's decay window to complete, the service returned a **retryable 5xx (503)** — prompting pointless retries for a request that can never succeed. This changes it to a **final HTTP 400** with a clear, actionable message.

## Root cause

In `lib/handlers/hard-quote/handler.ts`, `assertV2DecayWithinDeadline()` is invoked from `getCosignerData` / `getDefaultV2CosignerData`, which run **outside** the handler's `try/catch`. It threw a plain `Error`. The base API handler (`api-handler.ts`) only maps `CustomError` instances to a status code — everything else falls through to `INTERNAL_ERROR` (5xx), surfacing to the customer (via TAPI) as a retryable 503.

This check is V2-only by design: V3 allows partial decay past the deadline (see handler.ts comment), so it has no equivalent check.

## Changes

- **`lib/util/errors.ts`** — New `OrderDeadlineExpired` `CustomError` → HTTP **400** / `VALIDATION_ERROR`, with message: _"Order deadline is too close or has already expired; the order can no longer be filled. Resubmit the order with a later deadline."_
- **`lib/handlers/hard-quote/handler.ts`** — `assertV2DecayWithinDeadline` now throws `OrderDeadlineExpired` instead of a plain `Error`.
- **Tests**
  - `handlerDutchV2.deadline.test.ts` (existing unit test) updated to assert the typed error + message.
  - `handlerDeadline.test.ts` (new) — end-to-end handler tests asserting the customer-facing **400** on both the RFQ and open-order paths, plus a 200 happy-path guard.

## Testing

- Full unit suite green (168 passed): `jest --testPathIgnorePatterns test/integ dist/`
- Typecheck (`tsc --noEmit`) and eslint clean on changed files.
- Verified no downstream consumers pattern-match on the old error message string.

## Note (not in scope)

`test/handlers/hard-quote/handler.test.ts` contains an `it.only('No quotes')` (committed 2024-03) that silently disables the other 10 tests in that file in CI; 5 of them fail on stale expectations unrelated to this change (e.g. `HardQuoteRequest` now derives `requestId` from `quoteId ?? uuidv4()`). Left untouched here to keep this PR focused — worth a follow-up cleanup since that handler currently has minimal effective CI coverage.

### DIFF
--- a/lib/util/errors.ts
+++ b/lib/util/errors.ts
@@ -54,7 +54,7 @@ export class OrderPostError extends CustomError {
 
 export class OrderDeadlineExpired extends CustomError {
   private static MESSAGE =
-    'Order deadline is too close or has already expired; the order can no longer be filled. Resubmit the order with a later deadline.';
+    'Order deadline is too close or has already expired; the order can no longer be filled. Recreate the order with a later deadline.';
 
   constructor(message?: string) {
     super(message ?? OrderDeadlineExpired.MESSAGE);
--- a/lib/util/errors.ts
+++ b/lib/util/errors.ts
@@ -52,6 +52,28 @@ export class OrderPostError extends CustomError {
   }
 }
 
+export class OrderDeadlineExpired extends CustomError {
+  private static MESSAGE =
+    'Order deadline is too close or has already expired; the order can no longer be filled. Resubmit the order with a later deadline.';
+
+  constructor(message?: string) {
+    super(message ?? OrderDeadlineExpired.MESSAGE);
+    // Set the prototype explicitly.
+    Object.setPrototypeOf(this, OrderDeadlineExpired.prototype);
+  }
+
+  toJSON(id?: string): APIGatewayProxyResult {
+    return {
+      statusCode: 400,
+      body: JSON.stringify({
+        errorCode: ErrorCode.ValidationError,
+        detail: this.message,
+        id,
+      }),
+    };
+  }
+}
+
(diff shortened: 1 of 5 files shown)


lib/util/errors.ts
```diff
@@ -52,6 +52,28 @@ export class OrderPostError extends CustomError {
   }
 }
 
+export class OrderDeadlineExpired extends CustomError {
```
codyborn, line 55: Because it's a CustomError, it'll get returned in the response:
```
catch (err) {
            if (err instanceof CustomError) { // <-- enters here
              const errorJson = err.toJSON(id);
              log.error({ errorJson }, 'Unexpected error in handler');
              return errorJson;
            }
            log.error({ err }, 'Unexpected error in handler');
            metric.putMetric(Metric.QUOTE_500, 1, MetricLoggerUnit.Count);
            return INTERNAL_ERROR(id);
          }
```

ericneil-sanc (approved):
Seems good, only thing to note would be that I don't think the new Error wont show up as a metric anywhere, not sure if we want to track it anywhere, since we tracked it before

PR #445, opened 2026-06-10
feat: add ProtocolVersion.V3 and derive protocol from order type in HardQuoteRequest

codyborn:
## Summary

The hard-quote handler already parses and cosigns UniswapX DutchV3 orders, but the RFQ webhook side had no notion of a V3 protocol version: the `ProtocolVersion` enum stopped at V2, and `HardQuoteRequest.toCleanJSON()` hardcoded `protocol: 'v2'` even when the inner order was a DutchV3 order. As a result, V3 hard quotes were mislabeled as `v2` and routed using V2 support filters, and fillers had no way to opt into V3 RFQs.

## Changes

- **`lib/providers/webhook/index.ts`** — add `V3 = 'v3'` to the `ProtocolVersion` enum. This automatically flows into the Joi `FieldValidator.protocol` (validated against `Object.values(ProtocolVersion)`), so the indicative-quote API accepts `protocol: "v3"` and webhook configs can declare `supportedVersions: ['v3']`.
- **`lib/entities/HardQuoteRequest.ts`** — replace the hardcoded `protocol: ProtocolVersion.V2` in `toCleanJSON()` with a `protocol` getter backed by a `private readonly _protocol` field set in the constructor's existing orderType branch (`Dutch_V2` → `v2`, `Dutch_V3` → `v3`).
- **`lib/quoters/WebhookQuoter.ts`** — `getEndpointSupportedProtocols` now defaults unconfigured fillers to `[v2, v3]` (previously `[v2]`), so v3 hard quotes get RFQ pricing without each filler having to update their webhook config first. Fillers with an explicit `supportedVersions` list are unaffected and must include `v3` to receive V3 RFQs.

## Testing

- `test/entities/HardQuoteRequest.test.ts`: new V3 order fixture (`V3DutchOrderBuilder`, Arbitrum) with tests for the `protocol` getter and `toCleanJSON`/`toOpposingCleanJSON` emitting `v3`.
- `test/providers/quoters/WebhookQuoter.test.ts`: new test that a `v3` quote request is sent to fillers declaring V3 support and to default-config fillers, but not to fillers with an explicit V1/V2-only list.
- `tsc --noEmit` clean; all 25 unit test suites pass (integ suites require a deployed `UNISWAP_API` and were not run).

## Rollout note

Because the default now includes `v3`, fillers that never set `supportedVersions` will start receiving DutchV3 RFQs (block-based decay, 0.25% exclusivity override) as soon as this deploys. Fillers with an explicit `supportedVersions` config must add `v3` to participate.

### DIFF
--- a/lib/entities/HardQuoteRequest.ts
+++ b/lib/entities/HardQuoteRequest.ts
@@ -41,7 +41,7 @@ export class HardQuoteRequest {
       type: TradeType[this.type],
       numOutputs: this.numOutputs,
       ...(this.quoteId && { quoteId: this.quoteId }),
-      protocol: ProtocolVersion.V2,
+      protocol: this.protocol,
     };
   }
 
@@ -135,6 +135,16 @@ export class HardQuoteRequest {
 
   public get numOutputs(): number {
     return this.order.info.outputs.length;
+  }
+
+  public get protocol(): ProtocolVersion {
+    if (this.order instanceof UnsignedV2DutchOrder) {
+      return ProtocolVersion.V2;
+    } else if (this.order instanceof UnsignedV3DutchOrder) {
+      return ProtocolVersion.V3;
+    } else {
+      throw new Error('Unsupported order type');
+    }
   }
 
   public get cosigner(): string {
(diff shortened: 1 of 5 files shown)


ericneil-sanc (commented):
getEndpointSupportedProtocols will disqualify all filles by default unless they have been explicitly opted into v3 for "supportedVersion" in the config, which we have not been doing.

Should probably change it to return `[ProtocolVersion.V2, ProtocolVersion.V3]` instead of `[ProtocolVersion]`

alanhwu:
May be cleaner to do a private readonly `_protocol` var in the class and set it in the constructor in the `orderType` check we do there.

so that the `protocol()` getter is just returning `this._protocol`

alanhwu (approved):

PR #446, opened 2026-06-11
fix: raise default webhook timeout from 250ms to 500ms on all chains

codyborn:
## Summary
- Bump `WEBHOOK_TIMEOUT_MS_DEFAULT` from 250ms to 500ms, matching the existing mainnet timeout
- Update `getWebhookTimeoutMs` tests to expect 500ms on all chains

## Motivation
Quote requests on non-mainnet chains (observed on BNB chain 56) were returning `404 No quotes available` — the full webhook fan-out resolved at exactly the 250ms timeout boundary, meaning fillers were being cut off by the axios timeout before they could respond. 250ms proved too tight for filler RFQ latency, so this aligns all chains with the 500ms mainnet setting. Per-filler `overrides.timeout` config still takes precedence.

## Testing
- `yarn jest test/util/constants.test.ts` — 9/9 passing

### DIFF
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -38,10 +38,10 @@ export const POST_ORDER_ERROR_REASON = {
   INSUFFICIENT_FUNDS: 'Onchain validation failed: InsufficientFunds',
 };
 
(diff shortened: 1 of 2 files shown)


dannythedawger (approved):
lgtm

PR #450, opened 2026-06-12
fix: reconcile order-post timeouts instead of reporting rejection (SWAP-2839)

codyborn:
## Context

[SWAP-2839](https://linear.app/uniswap/issue/SWAP-2839/uniswapx-order-timeouts-can-still-create-filled-orders): swappers intermittently get `400 InputValidationError` / `timeout of 2000ms exceeded` from TAPI while the order is actually accepted and later filled on-chain.

Root cause (confirmed via Datadog trace `2119114652437636364`, mainnet DUTCH_V2): **this service's** `UniswapXServiceProvider` posts the cosigned order to uniswapx-service with a 2s axios timeout. The order service Lambda has a 29s budget and keeps running after we hang up — it persists the order and fillers fill it. The hard-quote handler then rewrote the timeout into a **400**, which TAPI surfaces to clients as `InputValidationError`. By the time we report "rejected", the order is live and fillable.

## Changes

1. `ORDER_SERVICE_TIMEOUT_MS` 2s → 7s. The order service validates on-chain (RPC `eth_call`) before accepting; 2s left no headroom. 7s stays under the hard-quote Lambda's 30s budget and under TAPI's hard-quote client timeout (raised in the companion TAPI PR).
2. **Reconcile on timeout**: a timeout is indeterminate, not a rejection. On timeout/connection-drop we now fetch the order by hash (`GET dutch-auction/orders?orderHash=`); if it was accepted we return success, so the client receives the cosigned order instead of a bogus rejection.
3. **Status mapping**: if reconciliation can't find the order, return **500 (status unknown)** — and the hard-quote handler no longer rewrites 5xx/indeterminate outcomes to 400. Only genuine 4xx rejections from the order service surface as 400.

Residual race: if the order service is still mid-request when we reconcile, the order can persist after our check — the window is now bounded by the companion uniswapx-service PR (Uniswap/uniswapx-service#682), which caps its RPC validation at 5s so its worst case fits inside our 7s.

## Testing

- New unit tests: `test/providers/order/uniswapxService.test.ts` (success, genuine 400 passthrough, timeout+reconcile-found → success, timeout+not-found → 500, reconcile failure → 500) and `test/handlers/hard-quote/handlerPostError.test.ts` (201→200, 400→400, 500→500).
- Full unit suite: 188 passed, 0 failed (skips pre-existing).
- Note: `test/handlers/hard-quote/handler.test.ts` has a stray `it.only` (from #438) that skips most of that file, and several of those hidden tests fail on main when un-skipped. Left untouched here — worth a separate fix.

### DIFF
--- a/lib/providers/order/uniswapxService.ts
+++ b/lib/providers/order/uniswapxService.ts
@@ -77,13 +77,14 @@ export class UniswapXServiceProvider implements OrderServiceProvider {
           }
           this.log.error({ orderHash, error: e.message }, 'Order post timed out and order was not found');
           // Status is genuinely unknown: the order may still be accepted and
-          // filled. Return the hash so the caller can reconcile it rather than
-          // treat the order as rejected (SWAP-2839).
+          // filled. Return the hash (same { hash } shape as a success) so the
+          // caller can reconcile it rather than treat the order as rejected
+          // (SWAP-2839).
           return {
             statusCode: 500,
             errorCode: ErrorCode.InternalError,
             detail: 'Timed out posting order to UniswapX Service; order acceptance could not be confirmed',
-            orderHash,
+            data: { hash: orderHash },
           };
         }
         this.log.error({ error: e.response?.data, httpStatus: e.response?.status, code: e.code }, 'Error posting order to UniswapX Service');
--- a/lib/providers/order/uniswapxService.ts
+++ b/lib/providers/order/uniswapxService.ts
@@ -6,7 +6,15 @@ import { ErrorResponse } from '../../handlers/base';
 import { ErrorCode } from '../../util/errors';
 import { CosignedV2DutchOrder, CosignedV3DutchOrder, OrderType } from '@uniswap/uniswapx-sdk';
 
-const ORDER_SERVICE_TIMEOUT_MS = 2000;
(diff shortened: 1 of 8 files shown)


lib/providers/order/uniswapxService.ts
```diff
@@ -67,4 +103,19 @@ export class UniswapXServiceProvider implements OrderServiceProvider {
       }
     }
   }
+
+  private async orderExists(chainId: number, orderHash: string): Promise<boolean> {
+    try {
+      // Give the order service's in-flight request a moment to persist.
+      await new Promise((resolve) => setTimeout(resolve, ORDER_RECONCILE_DELAY_MS));
+      const response = await axios.get(`${this.uniswapxServiceUrl}dutch-auction/orders`, {
```
alanhwu: Could be clean to have a type for this response type

alanhwu (approved):

lib/providers/order/uniswapxService.ts
codyborn: Good call — added a `GetOrdersResponse` type for the reconcile lookup and typed the `axios.get` generic, in 7db9154.

PR #454, opened 2026-06-17
feat: trigger circuit breaker on Laplace-smoothed fade rate

codyborn:
## Why

The V2 circuit breaker blocked a filler on **any** new fade within a ~10-minute window — an absolute-count trigger that structurally over-penalizes high-volume fillers (more orders won → more absolute fades, even at a healthy fade *rate*). This reworks it into a volume-aware, rate-based breaker, then hardens the surrounding state model.

## Trigger — Laplace-smoothed fade rate

```
fadeRate = (fades + α) / (total + α + β)      // α=1, β=19 → denominator (total + 20), prior mean = 1/20 = 5%
block when fadeRate > 12%                       // FADE_RATE_BLOCK_THRESHOLD
```

Smoothing seeds every filler with 1 pretend-fade / 19 pretend-clean-fills, pulling small samples toward the 5% prior so a couple of early fades don't trip the breaker, while a sustained high rate on real volume does. (Threshold must exceed the prior mean, else the prior alone blocks everyone — 12% > 5%. ✓)

## Measurement window — 24h, capped at the latest 100 orders/filler

The rate is computed over orders whose `deadline` falls in the last **24 hours**, capped at each filler's **latest 100 orders** (`ORDERS_PER_FILLER_LIMIT`). The cap makes the lookback adaptive, closing the evasion cases raised in review:

| scenario | smoothed rate | blocked? |
|---|---|---|
| low-volume filler fading every order (1/hr) | 2 fades → 3/22 = **13.6%** | ✅ (within ~2h) |
| high-volume chronic ~14% fader (7 of 50/hr) | latest 100 ≈ 14 fades → 15/120 = **12.5%** | ✅ |
| acute failure after clean history | ~12 fades to trip | ✅ (~1–2 cron cycles) |
| legitimate sustained ~10% filler | 11/120 = 9.2% | ❌ (correctly clear) |

In-flight (not-yet-completed) orders are filtered out **before** the per-filler `ROW_NUMBER()` cap, so they can't consume latest-N slots and skew a busy filler's sample.

## Recovery — timer-driven, with a clean slate

A blocked filler receives no RFQs, so it can't improve its rate while benched — recovery is driven by the **block timer**, not the rate. When the block expires the filler is eligible again and is scored on a **clean slate**: only orders completed after the block ended count, via a dedicated `fadeWindowStart` floor (see schema below). `consecutiveBlocks` escalation persists across blocks and decays gradually — but **only on runs where the filler actually completed new orders**, so a filler can't work off escalation by simply idling while stale rows linger in the 24h view.

## In-flight fades while blocked (`duringBlockRate`)

Orders that completed during a block are scored as their own smoothed cohort against the same 12% threshold. This **extends** an active block when the in-flight cohort genuinely faded (rate-based, so a high-volume filler's stray in-flight fade offset by clean fills does *not* extend), and **re-blocks right after expiry** when late in-flight fades landed between the last blocked-state cron run and `blockUntil` — closing a slip-through where such fades were scored by neither path.

Deliberately accepted (documented in code): because the window is a rolling 24h, a filler's rate can cross the threshold on a run with no new completions ("blocked while idle"). Gating the block on new completions would just move the surprise to right after a completed order, which is stranger UX.

## Escalation / block duration

Block length backs off purely on `consecutiveBlocks` (`2^n`, base 15 min); the old per-fade `1.2^(n-1)` multiplier is dropped so block *length* no longer scales with absolute fade count either.

## Fresh V2 state table + schema cleanup

State moves to a **new DynamoDB table (`FillerCBTimestampsV2`), created alongside the V1 table** rather than renaming it. The circuit-breaker state is derived (recomputed each cron run from Redshift), so the new table starts empty — no migration. This gives a **clean rollback**: the table name is compiled into each code version, so reverting to the old code automatically points back at the untouched V1 table, and the new version's writes never contaminate it. (Remove the V1 table in a follow-up once V2 is confirmed stable.)

The fresh table also lets us fix the stored schema, not just the name:
- **Native numbers, not strings.** All fields were `type:'string'` read via `parseInt`; with `wrapNumbers` this was the entire source of the NaN-guard complexity. Now stored as numbers and read via a `wrapNumbers:false` client the repo builds itself — a missing attribute is `undefined` (→ sentinel), never `NaN`.
- **De-overloaded `blockUntilTimestamp`.** It previously meant both "block expiry" and "clean-slate floor". Split out **`fadeWindowStart`** as the floor; `blockUntilTimestamp` now resets cleanly to `0` on decay while `fadeWindowStart` independently holds the last block end.
- **`lastPostTimestamp` → `lastExaminedTimestamp`** (it's the last cron-run boundary, not a post time).

## Row cap

The fade view's `LIMIT` was raised 5000 → **20000** (`FADE_QUERY_ROW_LIMIT`); the 1h→24h widening multiplied row volume ~24×, and above the cap rows are silently truncated (dropping fillers from evaluation). 20000 is ~200 filler-addresses of headroom at the 100/filler cap.

## API changes
- `getFillersNewFades` → `getFillersFadeStats`, returning `{ fadeRate, duringBlockRate, newCompletions }`.
- `calculateBlockUntilTimestamp(fromTimestamp, consecutiveBlocks)` (dropped the `fades` arg).
- `TimestampRepoRow`: `lastPostTimestamp`→`lastExaminedTimestamp`, `+fadeWindowStart`, all numbers.

## Tests
Rewrote `fade-rate-v2.test.ts` and updated the repo/provider/fixtures for the new model: smoothing math, windowing + clean-slate exclusion, address aggregation, backoff, all state transitions (block / decay+preserve-floor / extend-while-blocked / keep-while-blocked / re-block-after-expiry), anti-gaming decay + idle freeze, and the evasion scenarios pinned to the window params. Build + eslint clean; full unit suite passing.

## Rollout / rollback
- Deploy provisions the empty V2 table; the breaker repopulates within a cron cycle or two.
- Rollback = redeploy the previous version → automatically reads the untouched V1 table. No migration, no cleanup.

## Follow-ups (not in this PR)
- Observability (metrics + dashboard) is stacked in #460.
- Threshold / `α,β` / window should be **calibrated against the real per-filler `(fades, total)` distribution** before rollout.

### DIFF
--- a/lib/cron/fade-rate-v2.ts
+++ b/lib/cron/fade-rate-v2.ts
@@ -121,6 +121,14 @@ async function main(metrics: MetricsLogger) {
   }
 }
 
+// parseInt on a missing/malformed Dynamo attribute yields NaN. A NaN timestamp must not
+// be treated as a real value: e.g. `deadline > NaN` is always false, which would silently
+// zero out the fade-rate window and make a filler permanently unblockable. Coerce any
+// non-finite stored timestamp back to the "unset" sentinel.
+function finiteOr(value: number | undefined, fallback: number): number {
+  return Number.isFinite(value) ? (value as number) : fallback;
+}
+
 function newConsecutiveBlocks(consecutiveBlocks?: number): number {
   if (!consecutiveBlocks) {
     return 1;
@@ -213,7 +221,7 @@ export function calculateNewTimestamps(
       updatedTimestamps.push({
         hash,
         lastPostTimestamp: newPostTimestamp,
-        blockUntilTimestamp: fillerTimestamp?.blockUntilTimestamp ?? UNBLOCKED_BLOCK_UNTIL_TIMESTAMP,
+        blockUntilTimestamp: finiteOr(fillerTimestamp?.blockUntilTimestamp, UNBLOCKED_BLOCK_UNTIL_TIMESTAMP),
         consecutiveBlocks: decayedBlocks,
       });
     }
@@ -274,13 +282,14 @@ export function getFillersFadeStats(
       tallies[fillerHash] = { windowFades: 0, windowTotal: 0, newFades: 0 };
     }
     // Rate window: orders completed after the filler's last block ended (clean slate).
-    const windowStart = fillerTimestamp?.blockUntilTimestamp ?? UNBLOCKED_BLOCK_UNTIL_TIMESTAMP;
+    const windowStart = finiteOr(fillerTimestamp?.blockUntilTimestamp, UNBLOCKED_BLOCK_UNTIL_TIMESTAMP);
     if (row.deadline > windowStart) {
       tallies[fillerHash].windowTotal += 1;
       tallies[fillerHash].windowFades += row.faded;
     }
     // New fades since the last cron run (deadline-based; catches in-flight orders).
-    if (!fillerTimestamp || row.deadline > fillerTimestamp.lastPostTimestamp) {
+    const lastPostTimestamp = finiteOr(fillerTimestamp?.lastPostTimestamp, 0);
+    if (!fillerTimestamp || row.deadline > lastPostTimestamp) {
       tallies[fillerHash].newFades += row.faded;
     }
   });
--- a/lib/cron/fade-rate-v2.ts
+++ b/lib/cron/fade-rate-v2.ts
@@ -34,11 +34,11 @@ export const BASE_BLOCK_SECS = 60 * 15; // 15 minutes
 
 // Laplace (additive) smoothing applied to each filler's fade rate so a few fades on a
 // small sample don't trip the breaker. Equivalent to seeding every filler with ALPHA
-// pretend-fades and BETA pretend-clean-fills. Prior mean = ALPHA/(ALPHA+BETA) ≈ 4.8%.
+// pretend-fades and BETA pretend-clean-fills. Prior mean = ALPHA/(ALPHA+BETA) = 1/20 = 5%.
 export const LAPLACE_ALPHA = 1;
-export const LAPLACE_BETA = 20;
+export const LAPLACE_BETA = 19;
 // Block a filler once their smoothed fade rate exceeds this. MUST be greater than the
-// prior mean (~4.8%), otherwise the prior alone would block every filler.
+// prior mean (5%), otherwise the prior alone would block every filler.
 // At this threshold a filler needs e.g. ~3 fades in 10 orders, ~4 in 20, ~8 in 50.
 export const FADE_RATE_BLOCK_THRESHOLD = 0.12;
 
--- a/lib/cron/fade-rate-v2.ts
+++ b/lib/cron/fade-rate-v2.ts
@@ -21,11 +21,26 @@ import { DynamoFillerAddressRepository } from '../repositories/filler-address-re
 import { TimestampRepository } from '../repositories/timestamp-repository';
 import { STAGE } from '../util/stage';
 
-export type FillerFades = Record<string, number>;
+export type FillerFadeStats = {
+  // Laplace-smoothed fade rate over the filler's post-block window (see getFillersFadeStats).
+  fadeRate: number;
+  // Orders that faded since the last cron run; used to stack penalties while already blocked.
+  newFades: number;
+};
+export type FillerFadeStatsMap = Record<string, FillerFadeStats>;
 export type FillerTimestamps = Map<string, Omit<TimestampRepoRow, 'hash'>>;
 
 export const BASE_BLOCK_SECS = 60 * 15; // 15 minutes
-export const NUM_FADES_MULTIPLIER = 1.2;
+
+// Laplace (additive) smoothing applied to each filler's fade rate so a few fades on a
+// small sample don't trip the breaker. Equivalent to seeding every filler with ALPHA
+// pretend-fades and BETA pretend-clean-fills. Prior mean = ALPHA/(ALPHA+BETA) ≈ 4.8%.
+export const LAPLACE_ALPHA = 1;
+export const LAPLACE_BETA = 20;
+// Block a filler once their smoothed fade rate exceeds this. MUST be greater than the
+// prior mean (~4.8%), otherwise the prior alone would block every filler.
+// At this threshold a filler needs e.g. ~3 fades in 10 orders, ~4 in 20, ~8 in 50.
+export const FADE_RATE_BLOCK_THRESHOLD = 0.12;
 
 /** Sentinel when the filler has no active block (always < now for real unix seconds). Avoids equaling lastPostTimestamp, which could briefly read as blocked under clock skew. */
 export const UNBLOCKED_BLOCK_UNTIL_TIMESTAMP = 0;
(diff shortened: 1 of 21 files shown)


lib/cron/fade-rate-v2.ts
```diff
@@ -207,77 +230,93 @@ export function calculateNewTimestamps(
...
-      }
+      return;
     }
+    const fillerTimestamp = fillerTimestamps.get(fillerHash);
+    if (!tallies[fillerHash]) {
+      tallies[fillerHash] = { windowFades: 0, windowTotal: 0, newFades: 0 };
+    }
+    // Rate window: orders completed after the filler's last block ended (clean slate).
+    const windowStart = finiteOr(fillerTimestamp?.blockUntilTimestamp, UNBLOCKED_BLOCK_UNTIL_TIMESTAMP);
+    if (row.deadline > windowStart) {
```
gretzke, line 361: A filler can be blocked on a run with no new fades, because clean orders age out of the 1 hour window while fades remain

lib/cron/fade-rate-v2.ts
```diff
@@ -207,77 +230,93 @@ export function calculateNewTimestamps(
...
+      tallies[fillerHash].windowTotal += 1;
```
gretzke, line 363: open orders are counted towards the number of rows, so an active filler with lots of open orders can have their sample size decreased which could bias recent fades more heavily than they would actually be

lib/cron/fade-rate-v2.ts
```diff
@@ -118,38 +141,38 @@ function newConsecutiveBlocks(consecutiveBlocks?: number): number {
...
-    const fades = row[1];
+  Object.entries(fillerFadeStats).forEach(([hash, stats]) => {
+    const { fadeRate, newFades } = stats;
     const fillerTimestamp = fillerTimestamps.get(hash);
     const isCurrentlyBlocked = fillerTimestamp && fillerTimestamp.blockUntilTimestamp > newPostTimestamp;
 
-    if (isCurrentlyBlocked && fades) {
-      // Stack penalties while blocked
-      // Extend the block from current blockUntil, not from now
+    if (isCurrentlyBlocked && newFades > 0) {
```
gretzke: clean in-flight orders are not counted, so absolute fades still trigger blocks, which could disproportionately affect high-volume fillers even though the absolute fade count should no longer matter.

lib/cron/fade-rate-v2.ts
```diff
@@ -172,13 +195,11 @@ export function calculateNewTimestamps(
         blockUntilTimestamp: fillerTimestamp.blockUntilTimestamp,
         consecutiveBlocks: fillerTimestamp.consecutiveBlocks,
       });
-    } else if (fades) {
-      const blockUntilTimestamp = calculateBlockUntilTimestamp(
-        newPostTimestamp,
-        fillerTimestamp?.consecutiveBlocks,
-        fades
-      );
+    } else if (fadeRate > FADE_RATE_BLOCK_THRESHOLD) {
```
gretzke: just flagging that a filler could complete 1 order per hour and fading each order without ever being blocked. At 50 orders per hour a filler can fade 7 orders per hour indefinitely without being blocked.

lib/cron/fade-rate-v2.ts
```diff
@@ -21,11 +21,26 @@ import { DynamoFillerAddressRepository } from '../repositories/filler-address-re
...
+// small sample don't trip the breaker. Equivalent to seeding every filler with ALPHA
+// pretend-fades and BETA pretend-clean-fills. Prior mean = ALPHA/(ALPHA+BETA) = 1/20 = 5%.
+export const LAPLACE_ALPHA = 1;
+export const LAPLACE_BETA = 19;
+// Block a filler once their smoothed fade rate exceeds this. MUST be greater than the
+// prior mean (5%), otherwise the prior alone would block every filler.
+// At this threshold a filler needs e.g. ~3 fades in 10 orders, ~4 in 20, ~8 in 50.
+export const FADE_RATE_BLOCK_THRESHOLD = 0.12;
 
 /** Sentinel when the filler has no active block (always < now for real unix seconds). Avoids equaling lastPostTimestamp, which could briefly read as blocked under clock skew. */
```
gretzke: stale docstring

lib/cron/fade-rate-v2.ts
```diff
@@ -106,6 +121,14 @@ async function main(metrics: MetricsLogger) {
   }
 }
 
+// parseInt on a missing/malformed Dynamo attribute yields NaN. A NaN timestamp must not
+// be treated as a real value: e.g. `deadline > NaN` is always false, which would silently
+// zero out the fade-rate window and make a filler permanently unblockable. Coerce any
+// non-finite stored timestamp back to the "unset" sentinel.
+function finiteOr(value: number | undefined, fallback: number): number {
```
gretzke: probably cleaner to integrate this into the row parsing in the timestamp repository directly?

lib/cron/fade-rate-v2.ts
gretzke, line 361: edge case: if an in-flight order fades near the end of a block (after the last cron run that sees the filler as blocked but before blockUntil), it slips through both paths. it's excluded from the rate window since its deadline is before the window floor, and by the next cron run the block has expired so the newFades > 0 extend branch never fires

lib/cron/fade-rate-v2.ts
codyborn: This is still accurate right? `UNBLOCKED_BLOCK_UNTIL_TIMESTAMP` is used when there's no active block.

lib/cron/fade-rate-v2.ts
codyborn: Good point. We should consider new fades and new fills in the numerator/denominator respectively.

lib/cron/fade-rate-v2.ts
gretzke: so before this pr the decay branch always wrote the sentinel for every unblocked filler, now the decay preserves the old timestamp. so the "used when there's no active block" part is no longer true. if we set the sentinel for an unblocked filler we basically erase the new floor and score the entire history

lib/cron/fade-rate-v2.ts
codyborn: I've updated the logic to look at the last 100 fills over the past 24 hours. This could still happen, but much less likely. It'd require the filler to stop filling for a while.

codyborn: The during-block cohort is now timestamp-based (lastPost < deadline <= blockUntil), so late in-flight fades get scored on the next run and re-block even if the block expired in between (test: "re-blocks after expiry when in-flight orders faded during the block").

lib/cron/fade-rate-v2.ts
codyborn: you're right, updated the docstring in 5ccec64 to make clear the sentinel is only for never-blocked fillers and that recovered fillers intentionally keep their expired timestamp as the clean-slate floor.

lib/cron/fade-rate-v2.ts
codyborn: done in 68dbfde — parsing now coerces NaN to finite fallbacks in the repository itself, kept the cron-side guard as defense-in-depth for non-Dynamo sources.

lib/cron/fade-rate-v2.ts
codyborn: addressed in f135bb3 by widening the window to 24h / latest-100 orders — the 1/hr 100% fader now trips after 2 fades (3/22 ≈ 13.6%) and the 7-of-50/hr case lands at 15/120 = 12.5%.

lib/cron/fade-rate-v2.ts
codyborn: implemented in 0e21661 — the extend path now scores the in-flight cohort as a smoothed rate (duringBlockRate) against the same 12% threshold, so clean in-flight fills offset stray fades.

lib/cron/fade-rate-v2.ts
codyborn: The completed-orders filter now runs before ROW_NUMBER(), so open orders no longer eat cap slots.

gretzke:
consecutiveBlocks now decays on every cron run for any filler with rows in the 24h view, even if they completed nothing new. so a filler can reset their escalation by just going idle for ~30 min after a block instead of filling cleanly. before, fillers with no new orders kept their consecutiveBlocks frozen

codyborn:
true — fixed in acf0521: stats now report newCompletions (orders completed since the last run) and the decay branch only decrements consecutiveBlocks when it's > 0, so idling freezes escalation like the old code and working it off requires actual clean fills again.

gretzke (approved):

PR #456, opened 2026-06-18
fix: obfuscate two-sided RFQ quote to market makers

dannythedawger:
## What

We send both sides of every incoming RFQ to market makers — the real request plus an "opposing" (decoy) request — to obfuscate the trade direction. Today the two are trivially distinguishable, so the obfuscation buys ~no directional privacy:

- both POSTs to a filler carry the **same `requestId`**, explicitly stamping them as one synthetic pair;
- the **real request is always dispatched first**.

This PR closes the two safe, low-risk gaps — the first increment of SWAP-2914.

## Changes

- **Distinct decoy `requestId`** (`WebhookQuoter.fetchQuote`): the opposing request now gets its own `uuidv4()` on the wire, so a filler can't pair the two sides by `requestId`. Safe because the opposing response is matched in-memory (not by `requestId`) and `QuoteResponse.fromRFQ` does not validate `requestId` (only the real-path check does).
- **Randomized dispatch order**: which side is sent first is now random; the real-vs-opposing response mapping is tracked explicitly.

## Preserving bid/ask spread measurement

The two sides are still sent to the same filler, near-simultaneously, and paired in-memory. The opposing-side log is stamped with the **real** `requestId`, so any downstream join that computes the bid/ask spread by `requestId` keeps working even though the filler saw a distinct id on the wire.

## Out of scope (tracked on SWAP-2914)

- Sizing the decoy to a comparable notional via an indicative price instead of copying the amount.
- Symmetric framing so `type` (EXACT_INPUT vs EXACT_OUTPUT) no longer reveals which side is real.
- Productizing spread capture as a structured server-side event keyed by an internal pair id.

## Test plan

- `test/providers/quoters/WebhookQuoter.test.ts`: added coverage that (1) the opposing request is posted with a `requestId` distinct from the real one, (2) the real response is still selected when the opposing request is dispatched first, and (3) the opposing log retains the real `requestId` for spread correlation.
- Full unit suite green (183 passed); `tsc --noEmit` clean.

Fixes SWAP-2915

### DIFF
--- a/lib/quoters/WebhookQuoter.ts
+++ b/lib/quoters/WebhookQuoter.ts
@@ -146,10 +146,22 @@ export class WebhookQuoter implements Quoter {
     };
 
     try {
-      const [hookResponse, opposite] = await Promise.all([
-        axios.post(endpoint, cleanRequest, axiosConfig),
-        axios.post(endpoint, opposingCleanRequest, axiosConfig),
-      ]);
+      // The opposing request is sent on the wire with a DISTINCT requestId so a market maker
+      // cannot link it to the real request. Our logs keep the shared real requestId (via the
+      // child logger above) so the two legs stay correlated for debugging.
+      const opposingWireRequest = { ...opposingCleanRequest, requestId: uuidv4() };
+      // Randomize which side is dispatched first so the genuine request isn't deterministically
+      // sent ahead of the obfuscation request. The mapping back to real vs. opposing is tracked
+      // explicitly, so quote selection and spread logging are unaffected.
+      const realRequestFirst = Math.random() < 0.5;
+      const orderedRequests = realRequestFirst
+        ? [cleanRequest, opposingWireRequest]
+        : [opposingWireRequest, cleanRequest];
+      const [firstResponse, secondResponse] = await Promise.all(
+        orderedRequests.map((req) => axios.post(endpoint, req, axiosConfig))
+      );
+      const hookResponse = realRequestFirst ? firstResponse : secondResponse;
+      const opposite = realRequestFirst ? secondResponse : firstResponse;
 
       metric.putMetric(Metric.RFQ_RESPONSE_TIME, Date.now() - before, MetricLoggerUnit.Milliseconds);
       metric.putMetric(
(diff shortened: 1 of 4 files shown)


lib/quoters/WebhookQuoter.ts
```diff
@@ -146,10 +133,23 @@ export class WebhookQuoter implements Quoter {
...
 
     try {
-      const [hookResponse, opposite] = await Promise.all([
-        axios.post(endpoint, cleanRequest, axiosConfig),
-        axios.post(endpoint, opposingCleanRequest, axiosConfig),
-      ]);
+      // The opposing request is sent on the wire with a DISTINCT requestId so a market maker
+      // cannot link it to the real request. Our logs (above and below) keep the shared real
+      // requestId so the two legs stay correlated for debugging.
+      const opposingWireRequest = { ...opposingCleanRequest, requestId: uuidv4() };
```
codyborn: We need to update the cleanRequest.requestId too, right?

lib/quoters/WebhookQuoter.ts
```diff
@@ -293,10 +296,13 @@ export class WebhookQuoter implements Quoter {
...
       ) {
         opposingResponse.response.setFillerResponseLatencyMs(rawResponse.latencyMs);
-        log.info({
+        this.log.info({
           eventType: 'QuoteResponse',
           body: {
             ...opposingResponse.response.toLog(),
+            // Correlate the opposing (bid/ask) log to the genuine request internally, even
+            // though the opposing request was sent to the filler with a distinct requestId.
+            requestId: request.requestId,
```
codyborn: We should log the original and obfuscated one (for both opposing and real requests). This way if a partner asks us to track down a requestId, we'll have it.

codyborn: The `quote.toLog()` is going to log the wrong requestId
https://github.com/Uniswap/uniswapx-parameterization-api/blob/9711eb1503dd3cefe62bbd921ea60b314e0b585d/lib/handlers/quote/handler.ts#L131

codyborn (approved):

PR #462, opened 2026-07-11
fix: first-writer-wins filler-address ownership + per-filler address cap

codyborn:
## Why

The RFQ circuit breaker attributes fades to a filler (webhook endpoint) via an address→filler map that is populated from the **caller-supplied `response.filler`** on every OK quote (`WebhookQuoter` → `addNewAddressToFiller`). The security review on #454 flagged two holes in that registration path:

1. **Attribution poisoning** — `addNewAddressToFiller` did an unconditional, last-write-wins put, so a filler could claim an address owned by *another* filler. With the rate-based breaker, binding a high-volume clean competitor's address to your own hash dilutes your smoothed fade rate below the block threshold (and griefs the competitor). One valid-looking quote, no auction win needed.
2. **Unbounded squatting** — there was no cap on how many addresses a single filler could register (`$add` to a set, forever).

## Fix

Both enforced in `addNewAddressToFiller` (the sole writer of the map), mirrored in the mock:

- **First-writer-wins ownership:** an address belongs to whichever filler registered it first; a claim from a different filler is ignored (logged). A genuine address migration between endpoints needs a manual override.
- **Per-filler cap:** `MAX_FILLER_ADDRESSES = 3`; registrations beyond the cap are ignored (logged).

Both rejections are **safe no-ops** — the order still quotes and fills; the address just isn't added to the attribution map — so this cannot break quoting. Registration is already fire-and-forget from the quote path.

## How much can a quoter squat on now?
**3** (was unlimited), and it can no longer claim addresses owned by other fillers at all.

## Scope / caveats
- This caps the **attribution map**, which is what the circuit breaker reads. Fully bounding `postedorders.filler` diversity (the raw rows the fade SQL counts) additionally requires **cosign-time enforcement** — refusing to cosign an order for a filler's (N+1)th address — which touches the quote hot path and is left as a follow-up. Residual without it: a filler rotating to an over-cap address gets that address's orders *unattributed* (uncounted), i.e. a mild evasion rather than a dilution/poison — acceptable given the breaker treats unknown addresses as new identities anyway.
- `MAX_FILLER_ADDRESSES = 3` matches the value discussed in review, but should be **calibrated against the real per-endpoint address distribution** before we rely on it — a legit high-rotation filler above the cap would be under-monitored (safe direction, not a break). Easy one-constant tune.

## Tests
`filler-address-repository.test.ts`: first-writer-wins (a different filler's claim on an owned address is ignored, ownership unchanged) and the cap (only the first `MAX_FILLER_ADDRESSES` register; the over-cap address is attributed to no filler). Full unit suite passing (191).

## Related
Follow-up from the #454 security review (independent subsystem; branches off `main`). The cosign-time cap and a `FILLERS_EVALUATED` truncation alarm remain as further follow-ups.

### DIFF
--- a/lib/repositories/filler-address-repository.ts
+++ b/lib/repositories/filler-address-repository.ts
@@ -10,6 +10,13 @@ export type DynamoFillerToAddressRow = {
   addresses: string[];
 };
 
+// Max distinct on-chain addresses a single filler (webhook endpoint) may register. Bounds
+// Sybil breadth: without a cap a filler could register unbounded addresses to spread fades or
+// inflate the circuit-breaker's row set. Rejected registrations are a safe no-op (the order
+// still quotes/fills; the address just isn't attributed), so this can't break quoting.
+// NOTE: tune against the real per-endpoint address distribution before tightening.
+export const MAX_FILLER_ADDRESSES = 3;
+
 export interface FillerAddressRepository {
   getFillerAddresses(filler: string): Promise<string[] | undefined>;
   getFillerByAddress(address: string): Promise<string | undefined>;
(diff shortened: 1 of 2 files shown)


codyborn:
Folded into #454 (cherry-picked as commit 3c1f675 on the circuit-breaker branch). Closing this standalone PR.

PR #470, opened 2026-07-29
test: unskip and fix hard quote tests

alanhwu:
## Why

`test/handlers/hard-quote/handler.test.ts` had an `it.only` from 2024-03-14 (`8becf4db`). It silenced the rest of that file **and** all of `HardQuoteResponse.test.ts`, which imported `getOrder` from it — jest's `.only` scoping spans a cross-test-file import. 22 tests, dark for two years.

The cost wasn't the hidden tests. Six weeks later `e542951` changed `HardQuoteRequest` to derive `requestId` from `quoteId`, breaking 7 of the invisible tests. CI stayed green, so nobody learned.

## No production source changed

`requestId: _data.quoteId ?? uuidv4()` looks like a bug that discards the client's `requestId`. It isn't — it landed via reviewed PR #317 (which contains a *rejected* schema-level alternative), and it's load-bearing: `hardrequests` has no `quoteId` column, so this is the only way the indicative quoteId reaches Redshift. **The tests were stale, not the source.** Follow-up to add that column is filed separately.

## What changed

- Removed the `.only`; un-nested 2 tests buried inside an `it()` callback (they never registered — they pass unmodified)
- New `test/fixtures/hard-quote.ts`, absorbing **4 duplicate `getOrder` copies**. Kills both cross-test-file imports (`schema.test.ts` had a second one) and the duplicate suite execution
- Repaired `getCosignerData` tests. `'exact output quote worse'` was **passing for the wrong reason**: the function branches on `request.type` (derived from the order's input curve), not the `TradeType` passed to the mock quote, which is inert. Exact-output cases need an input-decaying order; added `expect(hardRequest.type)` guards so this can't drift silently again
- Made `QUOTE_ID`/`REQUEST_ID` distinct — they were byte-identical, so every "echoes the requestId" assertion was vacuous
- `.eslintrc`: zero-dependency `no-restricted-syntax` guard for `.only`/`fit`/`fdescribe` **and** nested-`it`. Jest has no config option that fails on focused tests, and `eslint-plugin-jest` has no nested-tests rule

## Verification

`yarn test:unit`: 12 failed / 252 → **0 failed, 242 passed, 1 skipped / 243**. Total drops because the duplicate execution is gone. `yarn build` and `yarn lint` unchanged (0 errors).

Green wasn't enough — the old suite was green too, so it's mutation-tested:
- revert the `requestId` derivation → **15 failures across 3 suites** (previously zero)
- plant an `it.only` → sibling file still runs 6/6 instead of going fully skipped

## Notes for review

- `HardQuoteResponse.test.ts:177` asserts `expect(x).not.toEqual(x)` across two `toLog()` calls. Not a typo — it pins the non-memoized `quoteId` getter, which mints a new uuid per access. Fine to drop; nothing else depends on it.
- Only **7** distinct tests were failing, not 12 — the cross-test import double-counted 5.
- `handlerDutchV3.test.ts` un-skipped: its only test was `it.skip`, so the suite was inert. The stated
  reason ("until V3 Order Service is ready") was wrong — it uses `MockOrderServiceProvider`. The real
  blocker was an unmocked `StaticJsonRpcProvider` defaulting to `localhost:8545`, which made the V3
  `getBlockNumber()` call fail and the handler 500. Now mocked like its sibling, with added assertions
  on `decayStartBlock` and V3's 25-bps exclusivity. **Suite is now 29/29, zero skips.**

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -30,6 +30,27 @@
     "@typescript-eslint/no-empty-interface": "off",
     "@typescript-eslint/ban-types": "warn",
     "@typescript-eslint/no-unused-vars": ["warn", { "argsIgnorePattern": "^_" }],
-    "@typescript-eslint/ban-ts-comment": "off"
+    "@typescript-eslint/ban-ts-comment": "off",
+    // An it.only sat in test/handlers/hard-quote/handler.test.ts from March 2024 to July 2026
+    // and silenced 22 tests. CI stayed green, so a later intentional behavior change
+    // (the requestId derivation) broke those tests with nobody noticing. These three
+    // selectors catch the whole class. no jest plugin needed -- jest itself has no config
+    // option that fails a run containing focused tests, and jest/no-nested-tests
+    // does not exist.
+    "no-restricted-syntax": [
+      "error",
+      {
+        "selector": "CallExpression[callee.object.name=/^(it|test|describe)$/][callee.property.name='only']",
+        "message": "Focused test (.only) silently skips every sibling test. Remove it before committing."
+      },
+      {
+        "selector": "CallExpression[callee.name=/^(fit|fdescribe)$/]",
+        "message": "Focused test (fit/fdescribe) silently skips every sibling test. Use it()/describe() instead."
+      },
+      {
+        "selector": "CallExpression[callee.name=/^(it|test)$/] CallExpression[callee.name=/^(it|test|describe)$/]",
+        "message": "A test block nested inside an it()/test() callback never registers as a test. Move it out to the enclosing describe()."
+      }
+    ]
   }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -31,12 +31,12 @@
     "@typescript-eslint/ban-types": "warn",
     "@typescript-eslint/no-unused-vars": ["warn", { "argsIgnorePattern": "^_" }],
     "@typescript-eslint/ban-ts-comment": "off",
-    // An it.only sat in test/handlers/hard-quote/handler.test.ts from March 2024 to July 2026
-    // and silenced 22 tests. CI stayed green, so a later intentional behavior change
-    // (the requestId derivation) broke those tests with nobody noticing. These three
-    // selectors catch the whole class. no jest plugin needed -- jest itself has no config
-    // option that fails a run containing focused tests, and jest/no-nested-tests
-    // does not exist.
+    // Two ways a test can silently stop running while CI still reports green: a focused test
+    // (.only/fit/fdescribe) skips all its siblings, and a test block nested inside an it()
+    // callback never registers at all. Both are lint-only problems -- jest has no config
+    // option that fails a run containing focused tests, and eslint-plugin-jest has no
+    // equivalent of the third selector, so core no-restricted-syntax covers both without
+    // adding a dependency.
     "no-restricted-syntax": [
       "error",
       {
--- a/test/entities/HardQuoteRequest.test.ts
+++ b/test/entities/HardQuoteRequest.test.ts
@@ -12,9 +12,10 @@ const RAW_AMOUNT = BigNumber.from('1000000');
 // held regardless of whether requestId carried the caller's id or the quoteId.
 const REQUEST_ID = 'b45c2d1e-7f30-4a92-8c65-1d8e4f2a9b03';
 const QUOTE_ID = 'a83f397c-8ef4-4801-a9b7-6e79155049f6';
-// HardQuoteRequest's constructor derives `requestId: _data.quoteId ?? uuidv4()` (e542951 /
-// PR #317), so every JSON view below reports the indicative QUOTE_ID as the requestId -- the
-// caller's REQUEST_ID is discarded. Named rather than inlined so the substitution is visible.
+// HardQuoteRequest's constructor derives `requestId: _data.quoteId ?? uuidv4()`, so every JSON
+// view below reports the indicative QUOTE_ID as the requestId -- the caller's REQUEST_ID is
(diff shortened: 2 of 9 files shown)


.eslintrc
```diff
@@ -30,6 +30,27 @@
     "@typescript-eslint/no-empty-interface": "off",
     "@typescript-eslint/ban-types": "warn",
     "@typescript-eslint/no-unused-vars": ["warn", { "argsIgnorePattern": "^_" }],
-    "@typescript-eslint/ban-ts-comment": "off"
+    "@typescript-eslint/ban-ts-comment": "off",
+    // An it.only sat in test/handlers/hard-quote/handler.test.ts from March 2024 to July 2026
```
codyborn: Nit: this comment block bakes in the incident narrative ("sat ... from March 2024 to July 2026", "22 tests") alongside the durable design rationale. The narrative will rot the moment someone reads this a year out with no way to verify it, and it does not help explain the current rule. Suggest trimming to just the ongoing facts: why these three selectors exist, and why a zero-dep `no-restricted-syntax` was chosen over `eslint-plugin-jest` (not currently a dependency). The incident detail belongs in the PR description, not the source.

test/entities/HardQuoteRequest.test.ts
```diff
@@ -1,54 +1,26 @@
...
 
-const NOW = Math.floor(new Date().getTime() / 1000);
 const RAW_AMOUNT = BigNumber.from('1000000');
-const REQUEST_ID = 'a83f397c-8ef4-4801-a9b7-6e79155049f6';
+// Deliberately DIFFERENT uuids. These were identical, so every toCleanJSON assertion below
+// held regardless of whether requestId carried the caller's id or the quoteId.
+const REQUEST_ID = 'b45c2d1e-7f30-4a92-8c65-1d8e4f2a9b03';
 const QUOTE_ID = 'a83f397c-8ef4-4801-a9b7-6e79155049f6';
+// HardQuoteRequest's constructor derives `requestId: _data.quoteId ?? uuidv4()` (e542951 /
+// PR #317), so every JSON view below reports the indicative QUOTE_ID as the requestId -- the
```
codyborn: Nit: citing the commit hash and PR number here (`e542951 / PR #317`) is point-in-time narrative — useful for `git blame` archaeology, not for a reader trying to understand the current contract. The load-bearing fact worth keeping is "requestId derives from quoteId because `hardrequests` has no quoteId column"; the commit/PR citation can be dropped.

test/handlers/hard-quote/handler.test.ts
```diff
@@ -269,7 +238,50 @@ describe('Quote handler', () => {
     });
   });
 
-  it.only('No quotes', async () => {
+  // Pins the requestId contract introduced by e542951 (PR #317): HardQuoteRequest's
```
codyborn: Same pattern as HardQuoteRequest.test.ts: "introduced by e542951 (PR #317)" is a point-in-time citation rather than an ongoing design fact. Keep the explanation of *why* the derivation exists and why `.not.toEqual` is load-bearing; drop the commit hash / PR number.

test/fixtures/hard-quote.ts
```diff
@@ -0,0 +1,113 @@
+// Shared V2 hard-quote order fixtures.
+//
+// WHY THIS FILE EXISTS: `getOrder` used to be exported from
+// test/handlers/hard-quote/handler.test.ts and `getOrderInfo` from
+// test/entities/HardQuoteRequest.test.ts, and imported across suites. Importing a *.test.ts
+// module runs its body inside the importer's jest context, so the imported file's
+// describe/it blocks re-register in the importer's suite. Two consequences:
+//   1. the imported suite executes twice (once on its own, once inside the importer), and
+//   2. a single `it.only` anywhere in either file silences BOTH files.
+// (2) is what hid 22 tests from March 2024 to July 2026. Never import from a *.test.ts file.
```
codyborn: Nit: "hid 22 tests from March 2024 to July 2026" is incident narrative, not a design fact — it will be stale/unverifiable the moment someone reads this file later. The durable part worth keeping as-is is the mechanism explanation and "Never import from a *.test.ts file"; suggest dropping the date/count and letting the PR description carry the incident story.

codyborn (approved):
approved with nits

.eslintrc
alanhwu: Agreed — trimmed to just the durable rationale in c5bb7a8: what the three selectors catch (focused tests, and test blocks nested in an `it()` callback, which never register), and why core `no-restricted-syntax` rather than adding `eslint-plugin-jest` (it has no equivalent of the third selector, and jest itself has no config option that fails a run containing focused tests). Dates/counts moved to the PR description.

test/entities/HardQuoteRequest.test.ts
alanhwu: Done in c5bb7a8 — dropped the `e542951 / PR #317` citation and kept the load-bearing reason you named: requestId derives from quoteId because `hardrequests` has no quoteId column, so that is how the indicative quoteId reaches Redshift.

test/handlers/hard-quote/handler.test.ts
alanhwu: Done in c5bb7a8 — same treatment: commit hash and PR number dropped, kept why the derivation exists and the note that `.not.toEqual` is the assertion that fails if the derivation is removed.

test/fixtures/hard-quote.ts
alanhwu: Done in c5bb7a8 — dropped the date and count, kept the re-registration mechanism (imported `*.test.ts` bodies run in the importer's jest context, so the suite runs twice and a `.only` in either file silences both) and the rule it implies. Incident story lives in the PR description.

PR #479, opened 2026-08-06
fix, test: stop writing client IPs to CloudWatch logs

codyborn:
## What

The API handler logged the raw `APIGatewayProxyEvent` when `getRequestInjected` threw. That event carries the client IP in several fields — `requestContext.identity.sourceIp` and the forwarding headers (`X-Forwarded-For`, `X-Real-IP`, `CF-Connecting-IP`) — so every hit of that path wrote an IP into CloudWatch.

```
lib/handlers/base/api-handler.ts:173  log.error({ err, event }, 'Unexpected error building request injected.')
```

Two things make this worth fixing here specifically:

- **No Lambda in this service sets `logRetention`** (zero hits across `bin/`), so those log groups take CloudWatch's default and **never expire**.
- The call is `error` level and the logger has no debug branch (`:132` is just `test ? FATAL+1 : INFO`), so it always fires in prod.

## How

Added `redactEvent()` and applied it at the call site. It's an **allowlist rather than a denylist**, so a new IP-bearing header can't leak in by default.

- **Kept:** `resource`, `path`, `httpMethod`, `pathParameters`, `queryStringParameters`, `body`, `isBase64Encoded`, and exactly six `requestContext` fields — `requestId`, `path`, `stage`, `resourcePath`, `httpMethod`, `requestTimeEpoch`.
- **Dropped:** `headers`, `multiValueHeaders`, and every other `requestContext` field. That includes `identity` (which carries `sourceIp`, `userAgent`, `caller`, `user`) and also `authorizer`, `apiId`, `extendedRequestId` and the domain fields. Dropping `authorizer` matters independently of IPs, since it can carry claims.

This brings the Lambda logs in line with the API Gateway access logs, which already set `ip: false` (`bin/stacks/api-stack.ts:71`).

## Context

Mirrors Uniswap/uniswapx-service#692 — same helper, same test, adapted to this repo's style. Came out of a review of the WAF IP rate limiting in `bin/stacks/api-stack.ts`, which is fine as-is: IP addresses are personal data under GDPR (Recital 30; CJEU *Breyer*, C-582/14), and rate limiting for abuse prevention is a standard legitimate-interest use. Indefinite IP retention in application logs is the harder part to justify, and it was inadvertent rather than deliberate.

Deliberately **out of scope**, flagged for follow-up:

- **Log retention.** No log group in this service has a retention policy. Adding `logRetention` across all 13 Lambdas is a larger infra change that deserves its own PR and its own review. The access log group at `api-stack.ts:61` also takes the CDK two-year default, though it holds no IPs given `ip: false`.
- `sampledRequestsEnabled: true` (×2) retains sampled requests including Source IP for 3 hours. Genuinely useful when debugging a misfiring WAF rule, so left on.
- **Request bodies are still logged.** For hard quotes that means `encodedInnerOrder` and `innerSig`, and since this error path fires inside `getRequestInjected` — before anything is persisted — the log entry can be the only copy of a signed order. Keeping the body still looks right for debuggability, but it's a deliberate call rather than a free one.

## Testing

New `test/handlers/base/redact-event.test.ts` asserts the exact output shape and serializes the result to confirm no IP (or `x-api-key`) survives — so it fails loudly if someone re-adds a field later.

Uses `toStrictEqual` rather than `toEqual`, since `toEqual` ignores keys whose value is `undefined` and would miss a re-added field absent from the fixture. Verified on the uniswapx-service twin: adding `multiValueQueryStringParameters` to `redactEvent` fails under `toStrictEqual` and passes under `toEqual`.

- `npx jest` (excluding `test/integ`) — 30 suites / 261 tests pass
- `npx tsc --noEmit` — clean
- `npx eslint` on both files — clean

### DIFF
--- a/lib/handlers/base/api-handler.ts
+++ b/lib/handlers/base/api-handler.ts
@@ -26,6 +26,32 @@ const INTERNAL_ERROR = (id?: string) => {
 
 export type APIGatewayProxyHandler = (event: APIGatewayProxyEvent, context: Context) => Promise<APIGatewayProxyResult>;
 
+/*
+ * The raw APIGatewayProxyEvent carries the client IP in several places: requestContext.identity.sourceIp
+ * and any of the forwarding headers (X-Forwarded-For, X-Real-IP, CF-Connecting-IP, ...). IP addresses are
+ * personal data, and our Lambda log groups have no retention policy set, so never log the raw event.
+ * This is an allowlist rather than a denylist so a new IP-bearing header can't leak in by default.
+ */
+export function redactEvent(event: APIGatewayProxyEvent) {
+  return {
+    resource: event.resource,
+    path: event.path,
+    httpMethod: event.httpMethod,
+    pathParameters: event.pathParameters,
+    queryStringParameters: event.queryStringParameters,
+    body: event.body,
+    isBase64Encoded: event.isBase64Encoded,
+    requestContext: {
+      requestId: event.requestContext?.requestId,
+      path: event.requestContext?.path,
+      stage: event.requestContext?.stage,
+      resourcePath: event.requestContext?.resourcePath,
+      httpMethod: event.requestContext?.httpMethod,
+      requestTimeEpoch: event.requestContext?.requestTimeEpoch,
+    },
+  };
+}
+
 export type ApiRInj = BaseRInj & { requestId: string };
 
 export type APIHandleRequestParams<CInj, RInj, ReqBody, ReqQueryParams> = BaseHandleRequestParams<
@@ -170,7 +196,7 @@ export abstract class APIGLambdaHandler<
               metric
             );
           } catch (err) {
-            log.error({ err, event }, 'Unexpected error building request injected.');
+            log.error({ err, event: redactEvent(event) }, 'Unexpected error building request injected.');
             metric.putMetric(Metric.QUOTE_500, 1, MetricLoggerUnit.Count);
             return INTERNAL_ERROR();
           }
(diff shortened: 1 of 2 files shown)


gretzke (approved):
Same review as on the uniswapx-service twin. I ran the new test file plus npx tsc --noEmit on the branch, both clean, and grepped the repo to confirm this is the only place the raw event gets logged. Two small things.

1. Consider toStrictEqual instead of toEqual in the shape test. toEqual ignores keys whose value is undefined, so the test only fails for re-added fields that also exist in the fixture.

2. Worth being aware that the kept body still carries encodedInnerOrder and innerSig for hard quotes, and this error path fires before anything is persisted, so the log can be the only copy of a signed order. Keeping the body for debuggability still seems reasonable, just flagging it since the description doesn't call it out.

codyborn:
Thanks — both taken.

1. **`toStrictEqual`** — done in c984b64. Verified the failure mode on the uniswapx-service twin: adding `multiValueQueryStringParameters` to `redactEvent` (absent from the fixture, so `undefined`) fails under `toStrictEqual` and passes under `toEqual`. Added a comment above the assertion so the choice doesn't get reverted as noise.

2. **`encodedInnerOrder` / `innerSig`** — good catch, and the pre-persistence point is the sharper half of it. Now called out explicitly in the description: the body is still kept for debuggability, but stated as a deliberate call rather than something the description quietly skipped.

Also mirrored the `requestContext` wording fix from the service PR — the helper keeps six named fields and drops the rest including `authorizer`, which is worth dropping independently of IPs.

PR #481, opened 2026-08-10
feat: dashboard vertical change markets, config detection

alanhwu:


lib/providers/webhook/s3.ts
```diff
@@ -43,7 +47,62 @@ export class S3WebhookConfigurationProvider implements WebhookConfigurationProvi
       })
     );
     const s3Body = checkDefined(s3Res.Body, 's3Res.Body is undefined');
+    const previousNames = new Set(this.endpoints.map((e) => e.name));
```
ericneil-sanc: This line re-opens the throw path that detectConfigChange's try/catch was added to close — it dereferences the previous fetch's unvalidated payload with a bare e.name, outside the guard.

ericneil-sanc (approved):

PR #482, opened 2026-08-10
fix: circuit-breaker decay leak; cap block backoff; chronic-rate watchlist

codyborn:
## Why

The V2 breaker's exponential backoff never engages against chronic **low-volume** faders. Decay required only `newCompletions > 0`, and `newCompletions` counts *faded* completions — so a fader whose fades land in separate cron runs gets this cycle:

1. fade #1 → window rate 2/21 = 9.5%, under threshold → run counts as "activity" → **decay −1**
2. fade #2 → 3/22 = 13.6% → **block, +1**

Net escalation per cycle: **zero**. `consecutiveBlocks` pins at ~0–2, every block stays 15–30 minutes, and a filler with a ~2h natural order gap loses almost no volume to the bench.

Confirmed in production (a filler fading ~94% of a week's orders sat at `consecutiveBlocks = 2` with an active block) **and quantified by replaying 2 weeks of real order history** (24,315 orders) through the cron logic: under the old code, **57% of all decay events granted to chronic fillers came from runs whose completions included a fade** — for the extreme faders, 100% did. Moderate chronic faders (24–39% fade mix) sat in near-perfect trip/decay equilibrium (e.g. 64 trips vs 63 decays), `consecutiveBlocks` never exceeding 2 for the entire fortnight.

## Fix 1 — decay requires sustained clean activity

Decay one escalation level only after **`CLEAN_RUNS_PER_DECAY` (6) consecutive clean runs**, tracked in a new `consecutiveCleanRuns` field on the CB state row:

- **clean run** = ≥1 newly classified completion past the clean-slate floor **and** 0 new fades (per new `newFades` stat)
- any new fade **resets** the streak — a sub-threshold fade is not recovery; in-flight fades while blocked also reset it
- an idle run (no completions) **freezes** the streak — idling neither builds nor breaks it
- fills served while benched earn nothing (deadline on/before the floor is excluded from streak credit)
- a block/extension restarts the streak at 0; completing a streak decays one level and restarts it

**Load-lag tolerance:** streak inputs are classified against a horizon trailing wall clock by `STREAK_FINALITY_LAG_SECS` (2h) — each run classifies deadlines in `(lastExamined − LAG, now − LAG]`, which partitions time across runs with no extra state. Redshift is batch-loaded hourly, so without this a late-loading fade never reset the streak and a not-yet-loaded fill read as a transient fade that reset it spuriously. Rows are classified exactly once, after their load state is final. (Blocking itself stays on the un-lagged rate window — only decay is deferred.)

At the 10-minute cron cadence, recovery costs ≥1 hour of demonstrated clean activity per level, making decay structurally slower than escalation (+1 per block event).

**Backtested impact** (replay, same fortnight, fade semantics matching live SQL): chronic >30% faders go from 17% time benched to **59%**, with fades reaching swappers dropping **~70%** (605 → 180); the 16–30% band goes from 1.7% to **27%** benched (246 → 128 allowed fades); collateral on legitimate ≤11% fillers stays ≤0.8% bench time with ≤2h worst sentences.

## Fix 2 — cap block backoff at 2^7 (32h per increment), counter included

Uncapped, the replay produced a **128h (5.3-day)** block. `calculateBlockUntilTimestamp` caps the duration exponent at `MAX_BLOCK_BACKOFF_EXPONENT = 7`, and `newConsecutiveBlocks` caps the **stored counter** at the same bound — so full recovery is bounded at 7 × 6 clean runs (~7h of clean activity) instead of growing with block history. Projected cost: +38 allowed fades per fortnight across the worst offenders (+21%); benefit: bounded worst-case sentence (~64h with stacked extensions) and automatic recovery from pathological stored state.

## Fix 3 — chronic-rate watchlist metric + dashboard widget

New `CIRCUIT_BREAKER_V2_CHRONIC_RATE`: each filler's **raw** fade rate over the entire query window — no clean-slate amnesty, no smoothing — emitted at a ≥10-order sample, with a dashboard widget annotated at the block threshold.

Rationale: the backtest exposed a low-volume ~20% fader living permanently **inside** the threshold's envelope (`maxFadesPerDay ≈ 0.12n + 1.4`, i.e. ~21% raw at 15 orders/day). Every candidate mechanism was tested against them — threshold 10%, weaker prior (α=0.5), raw-rate dual rules, a chronic decay gate, and an EWMA-scored breaker — and **each either fails to contain them or benches legitimate fillers heavily**. The correct move is visibility, not a trigger change: this metric makes such fillers a deliberate human decision.

## Not included (deliberately)

- **EWMA-scored breaker**: backtested; structurally worse exit dynamics for high-volume fillers (a bursty hour → day-scale bench for a legitimate filler, since the score exits multiplicatively) and weaker than the ratchet against sparse extreme faders. May revisit as shadow-mode metric.
- **Threshold/prior recalibration**: every variant tested sits inside the collateral/containment frontier of the current parameters.
- **Detection latency** (hourly Redshift loads → up to ~1h of unprotected acute fading): tracked separately. The streak's finality lag tolerates this for *decay*; trip latency is inherent to the data path.

## Tests / docs

- Repro-first: the decay-leak suite drives the real `getFillersFadeStats → calculateNewTimestamps` pipeline with the production profile and failed on the old code (decay 2→1; next block 2^1 instead of 2^2).
- Streak build/complete/reset/freeze, deferred + exactly-once lag classification, bench-fill exclusion, duration + counter cap pinning, chronic-rate stats (no-amnesty semantics), metric emission gating, repo round-trip + missing-attribute default. `fadeStats()`/`cbState()` factories keep test literals to the fields under test.
- CLAUDE.md documents the backtest extract SQL (fade semantics kept in sync with `V2_FADE_RATE_SQL` — thanks review for catching the #461 drift) and replay conventions.

## Migration

None. Existing `FillerCBTimestampsV2` rows lack the new attribute; reads default it to 0 and the next cron write populates it.

## Review updates (post-review commit cc7c6d6)

All six review findings addressed: lag-tolerant streak classification, bench fills excluded from streak credit, `consecutiveBlocks` counter capped, `consecutiveCleanRuns` required in the write type, extract SQL corrected to `fillTimeBlocks > 0`, and test factories. The backtest was re-run with corrected fade semantics; all figures above reflect the corrected data.

## Hardening (post-review commit 5e28393)

Follow-ups from the `/code-review` pass and the eviction-boundary analysis, with the backtest rerun modeling the **complete final behavior** (lagged streak classification fed only from window-surviving rows, bench-fill exclusion, both caps, read clamp):

- **Read-site clamp**: stored `consecutiveBlocks` is clamped into `[0, MAX_BLOCK_BACKOFF_EXPONENT]` where state is read, so legacy over-cap rows normalize on every path (keep/decay/extend/re-block and the metric) and corrupted negative values self-heal instead of yielding `2^negative` sub-base blocks.
- **Chronic watchlist noise**: tallies score only rows past the finality horizon (not-yet-loaded fills can no longer sawtooth the metric), and emission requires a 6% rate floor (`CHRONIC_RATE_EMISSION_FLOOR`) so per-filler series exist only for watch-worthy fillers.
- **`CIRCUIT_BREAKER_V2_SATURATED_ADDRESSES`**: fires when an address's latest-N window is full AND its oldest row is fresher than the finality horizon — the true "rows evicted before streak classification" condition. Merely sitting at the latest-N cap is the designed adaptive window and is not flagged (backtest: cap occupancy is near-constant for big fillers — 98.8% of runs — while true degradation occurred on **1.0% of runs, confined to the two busiest, never-escalated addresses** at 104–123% of the cap during peak bursts). The boundary (~50 orders/hour sustained per address) is documented on `STREAK_FINALITY_LAG_SECS`.
- Chronic widget retitled: the window is adaptive (24h / latest-100), not "24h".

**Final backtest posture** (2 weeks, 24,315 orders, corrected fade semantics, eviction modeled): chronic >30% faders **50% benched, 218 allowed fades** (vs 17% / 605 pre-#482); 16–30% band 26% benched, 135 allowed (vs 1.7% / 246); worst continuous sentence **64h** (vs 128h uncapped); collateral on ≤11% fillers unchanged at ≤0.9% bench / ≤2h. The review fixes cost no containment beyond the already-priced cap.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -16,7 +16,7 @@ data-eng-workflows/lib/spaces/uniswap_x/functions/uniswap_x_hourly_config/tables
 Table → YAML mapping (table names are lowercased in Redshift; YAML uses snake_case):
 
 | Redshift table   | Load schema YAML       |
-|------------------|------------------------|
+| ---------------- | ---------------------- |
 | `postedorders`   | `posted_orders.yaml`   |
 | `archivedorders` | `archived_orders.yaml` |
 | `rfqrequests`    | `rfq_requests.yaml`    |
@@ -26,7 +26,7 @@ Table → YAML mapping (table names are lowercased in Redshift; YAML uses snake_
 in the corresponding YAML.** The view/query column references are validated only at runtime
 against the live cluster — there is no compile-time or unit-test check — so a typo or a
 non-loaded column fails the cron in production (and a column that exists but is null for the
-relevant rows fails *silently*).
+relevant rows fails _silently_).
 
 ### The trap: "emitted" ≠ "loaded"
 
@@ -61,7 +61,11 @@ posted while a filler would have been benched as prevented (see PR #482 discussi
 full harness design, per-filler duty-cycle/allowed-fades metrics, and baseline numbers).
 
 Extract query (matches the breaker's fade semantics from `V2_FADE_RATE_SQL`, but with **no
-24h window, no latest-100 cap, and no row limit** — the replay applies windowing itself):
+24h window, no latest-100 cap, and no row limit** — the replay applies windowing itself).
+**Keep the `faded` CASE in sync with `V2_FADE_RATE_SQL`** — e.g. #461 changed Dutch_V3 to
+`fillTimeBlocks > 0` (a fill at the decay-start block is _not_ a fade); an extract using the
+old `>= 0` inflates V3 fade rates and mis-calibrates every knob. The raw columns are included
+so the replay can recompute `faded` locally if the semantics change again:
 
 ```sql
 SELECT
@@ -79,7 +83,7 @@ SELECT
     ao.tokenOut AS tokenOut,
     CASE
       WHEN ao.fillTimestamp IS NULL THEN 1
-      WHEN po.ordertype = 'Dutch_V3' AND ao.fillTimeBlocks >= 0 THEN 1
+      WHEN po.ordertype = 'Dutch_V3' AND ao.fillTimeBlocks > 0 THEN 1
       WHEN po.ordertype = 'Dutch_V2' AND po.starttime < ao.fillTimestamp THEN 1
       ELSE 0
     END AS faded
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -50,3 +50,58 @@ for the rows you care about.
    ```
 3. New columns must be added to the `data-eng-workflows` load YAML (and the table) **before**
    any SQL here references them.
+
+## Backtesting fade circuit-breaker changes against real order history
+
+Any change to the circuit-breaker knobs in `lib/cron/fade-rate-v2.ts` (threshold, Laplace
+prior, block backoff/cap, decay rules, windowing) should be replayed against real order data
+before shipping — the PR #482 backtest reversed two confidently-held design opinions, cheaply.
+Method: pull the extract below, then simulate 10-minute cron runs over it, treating orders
+posted while a filler would have been benched as prevented (see PR #482 discussion for the
+full harness design, per-filler duty-cycle/allowed-fades metrics, and baseline numbers).
+
+Extract query (matches the breaker's fade semantics from `V2_FADE_RATE_SQL`, but with **no
+24h window, no latest-100 cap, and no row limit** — the replay applies windowing itself):
+
+```sql
+SELECT
+    po.filler   AS rfqFiller,       -- quoted exclusive filler address
+    po.quoteid  AS quoteId,
+    po.chainid  AS chainId,
+    po.ordertype AS orderType,
+    po.createdat AS postTimestamp,  -- epoch secs
+    po.deadline AS deadline,        -- epoch secs, completion time
+    po.starttime AS decayStartTime,
+    ao.fillTimestamp AS fillTimestamp,
+    ao.fillTimeBlocks AS fillTimeBlocks,
+    ao.filler AS actualFiller,
+    ao.tokenIn AS tokenIn,
+    ao.tokenOut AS tokenOut,
+    CASE
+      WHEN ao.fillTimestamp IS NULL THEN 1
+      WHEN po.ordertype = 'Dutch_V3' AND ao.fillTimeBlocks >= 0 THEN 1
+      WHEN po.ordertype = 'Dutch_V2' AND po.starttime < ao.fillTimestamp THEN 1
+      ELSE 0
+    END AS faded
+FROM postedorders po
+LEFT OUTER JOIN archivedorders ao ON po.quoteid = ao.quoteid
+WHERE po.ordertype IN ('Dutch_V2', 'Dutch_V3')
+  AND po.quoteid IS NOT NULL
+  AND po.filler IS NOT NULL
+  AND po.filler != '0x0000000000000000000000000000000000000000'
+  AND po.chainid NOT IN (5, 8001, 420, 421613)
+  AND po.deadline < EXTRACT(EPOCH FROM GETDATE())                          -- completed only
+  AND po.deadline >= EXTRACT(EPOCH FROM (GETDATE() - INTERVAL '28 DAYS'))  -- replay window
+ORDER BY po.deadline ASC;
+```
+
+Post-processing the replay must do itself (deliberately not in the SQL):
+
+- **Filter permissioned-token orders** using `PERMISSIONED_TOKENS` from `@uniswap/uniswapx-sdk`
+  (the breaker excludes them); the SQL keeps `tokenIn`/`tokenOut` for this so the list can't
+  drift from the code.
+- **Aggregate addresses to fillers** with the `FillerAddress` DynamoDB table
+  (`aws dynamodb scan --table-name FillerAddress`) — the breaker scores per filler hash, not
+  per address. Per-address replay is a usable approximation but under-counts multi-address
+  fillers.
+- Dedupe on `quoteId` (the `archivedorders` join can rarely fan out).
--- a/lib/cron/fade-rate-v2.ts
+++ b/lib/cron/fade-rate-v2.ts
@@ -41,6 +41,9 @@ export type FillerFadeStats = {
   // working off escalation requires demonstrated activity, not just going idle while stale
   // rows linger in the 24h view.
   newCompletions: number;
+  // Fades among newCompletions. A run with any new fade is not a clean run: it cannot build
+  // the decay streak (and resets it), so a sub-threshold fade never counts as recovery.
+  newFades: number;
 };
 export type FillerFadeStatsMap = Record<string, FillerFadeStats>;
 export type FillerTimestamps = Map<string, Omit<TimestampRepoRow, 'hash'>>;
@@ -57,6 +60,14 @@ export const LAPLACE_BETA = 19;
 // a sustained ~14%+ raw fade rate (e.g. ~3 fades in 10, ~14 in 100), or ~2 fades at very low
 // volume.
 export const FADE_RATE_BLOCK_THRESHOLD = 0.12;
+// Escalation (consecutiveBlocks) decays one level only after this many consecutive clean
+// runs — cron runs with at least one new completion and zero new fades. A run with any new
+// fade resets the streak; an idle run freezes it. At the 10-minute cron cadence this makes
+// recovery cost >=1 hour of demonstrated clean activity per level, so decay is structurally
+// slower than escalation (+1 per block event). Without the streak, a chronic low-volume
+// fader whose fades land in separate cron runs (fade #1 under threshold -> "activity" decay,
+// fade #2 -> block) nets zero escalation per cycle and stays on 15-30 minute blocks forever.
+export const CLEAN_RUNS_PER_DECAY = 6;
 
 const log = Logger.createLogger({
   name: 'FadeRate',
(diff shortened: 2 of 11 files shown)


lib/cron/fade-rate-v2.ts
```diff
@@ -354,12 +423,25 @@ export function getFillersFadeStats(
...
+        chronicFades: 0,
+        chronicTotal: 0,
+      };
     }
     const windowStart = fillerTimestamp?.fadeWindowStart ?? UNBLOCKED_BLOCK_UNTIL_TIMESTAMP;
     const lastExaminedTimestamp = fillerTimestamp?.lastExaminedTimestamp ?? 0;
+    // Chronic (no-amnesty) view: every row in the query window, regardless of cohort.
+    tallies[fillerHash].chronicTotal += 1;
+    tallies[fillerHash].chronicFades += row.faded;
     if (row.deadline > lastExaminedTimestamp) {
```
gretzke: This classifies a row as new activity by comparing its deadline to a watermark that jumps to wall clock on every run, which makes the streak sensitive to Redshift load lag in both directions.

postedorders and archivedorders are batch loaded hourly while the cron runs every 10 minutes, so a row can first become visible with a deadline that is already behind the watermark. A fade that arrives late never counts as a `newFade` and never resets the streak, so a chronic fader can decay an escalation level while fading inside the window. The fade still moves `fadeRate`, only the streak reset misses it.

The same lag also cuts the other way. An order whose fill row has not loaded yet is faded per the view's LEFT JOIN (`fillTimestamp IS NULL`) on the first run after its deadline, which is exactly the run where it counts as a `newFade` and resets the streak. When the fill row lands, the order flips back to faded=0, but the watermark has moved past it and the correction never re-enters the tally. The smoothed rate tolerates these transients, the streak does not, one transient resets it to 0, so an honest high-volume filler may never decay.

I think the streak needs to tolerate load lag before this ships, either by classifying "new" against a watermark that trails wall clock by the worst-case load lag, or by deriving the streak from the window contents instead of incrementally.

lib/cron/fade-rate-v2.ts
```diff
@@ -245,21 +288,38 @@ export function calculateNewTimestamps(
...
+      // fadeWindowStart is left as-is: it holds the last block end as the clean-slate
+      // floor, so a returning filler is scored only on orders completed after their block
+      // ended.
+      let consecutiveCleanRuns: number;
+      consecutiveBlocks = previousBlocks;
+      if (previousBlocks === 0) {
+        consecutiveCleanRuns = 0; // nothing to work off; don't accumulate a stale streak
+      } else if (newFades > 0) {
+        consecutiveCleanRuns = 0;
+      } else if (newCompletions > 0) {
```
gretzke, line 362: When a block expires between two cron runs, the first unblocked run lands here with `newCompletions` that can consist entirely of during-bench fills (deadline after the watermark but on or before `fadeWindowStart`). If none of them faded, the streak increments. That contradicts the rule in the blocked branch above, that serving a bench is not recovery. Had the cron run a minute before expiry, the same fills would have earned nothing. Restricting the increment to rows with deadline past `fadeWindowStart` would close this.

CLAUDE.md
```diff
@@ -50,3 +50,58 @@ for the rows you care about.
...
+    po.deadline AS deadline,        -- epoch secs, completion time
+    po.starttime AS decayStartTime,
+    ao.fillTimestamp AS fillTimestamp,
+    ao.fillTimeBlocks AS fillTimeBlocks,
+    ao.filler AS actualFiller,
+    ao.tokenIn AS tokenIn,
+    ao.tokenOut AS tokenOut,
+    CASE
+      WHEN ao.fillTimestamp IS NULL THEN 1
+      WHEN po.ordertype = 'Dutch_V3' AND ao.fillTimeBlocks >= 0 THEN 1
```
gretzke: The doc says this extract matches the fade semantics of `V2_FADE_RATE_SQL`, but the live SQL uses `fillTimeBlocks > 0` since #461. With `>= 0` the replay counts fills at the decay start block as fades and calibrates the knobs against inflated V3 fade rates.

lib/cron/fade-rate-v2.ts
```diff
@@ -396,8 +481,9 @@ export function getFillersFadeStats(
     - 1 consecutive block:  15 * 2^1 = 30 minutes
     - 2 consecutive blocks: 15 * 2^2 = 60 minutes
     - 3 consecutive blocks: 15 * 2^3 = 120 minutes
+    - 7+ consecutive blocks: capped at 15 * 2^7 = 32 hours per increment
 */
 export function calculateBlockUntilTimestamp(fromTimestamp: number, consecutiveBlocks: number | undefined): number {
-  const blocks = consecutiveBlocks || 0;
+  const blocks = Math.min(consecutiveBlocks || 0, MAX_BLOCK_BACKOFF_EXPONENT);
```
gretzke, line 588: This caps the block duration, but the stored `consecutiveBlocks` itself still grows without bound. Full recovery costs `consecutiveBlocks * 6` clean runs, so a filler with a long block history stays one sub-threshold fade away from a 32h block long after reforming. Capping the counter in `newConsecutiveBlocks` would bound both the duration and the recovery time.

lib/repositories/base.ts
```diff
@@ -34,15 +34,23 @@ export type TimestampRepoRow = {
...
 export type DynamoTimestampRepoRow = TimestampRepoRow;
 
-export type ToUpdateTimestampRow = Omit<TimestampRepoRow, 'blockUntilTimestamp' | 'fadeWindowStart'> & {
+export type ToUpdateTimestampRow = Omit<
+  TimestampRepoRow,
+  'blockUntilTimestamp' | 'fadeWindowStart' | 'consecutiveCleanRuns'
+> & {
   blockUntilTimestamp?: number;
   fadeWindowStart?: number;
+  consecutiveCleanRuns?: number;
```
gretzke: With this optional and `updateTimestampsBatch` doing a full-item put, a future caller that omits the field compiles fine and silently wipes a filler's streak (the read path defaults it to 0). The current writer sets it in all branches, so this is latent, but making it required like `consecutiveBlocks` closes it off.

test/crons/fade-rate-v2.test.ts
```diff
@@ -187,47 +203,83 @@ describe('FadeRateV2 cron', () => {
...
+      expect(capped - now).toEqual(32 * 3600); // pin the intended ceiling in wall-clock terms
+    });
   });
 
   describe('calculateNewTimestamps', () => {
     it('blocks a filler whose fade rate exceeds the threshold', () => {
       const timestamps: FillerTimestamps = new Map();
-      const stats: FillerFadeStatsMap = { newBad: { fadeRate: 0.2, duringBlockRate: 0.05, newCompletions: 1 } };
+      const stats: FillerFadeStatsMap = {
+        newBad: {
```
gretzke: Nit: there are about 20 of these literals now, each spelling out all six fields. A small factory with defaults, like `fadeStats({ fadeRate: 0.2, newFades: 1 })`, would keep the fields each test cares about visible and absorb the next field addition.

lib/cron/fade-rate-v2.ts
codyborn: Confirmed and fixed in cc7c6d6 — this was the sharpest catch of the review; both directions were real. Streak inputs are now classified against a horizon trailing wall clock by `STREAK_FINALITY_LAG_SECS` (2h): each run classifies deadlines in `(lastExamined − LAG, now − LAG]`. Since `lastExaminedTimestamp` advances to `now` every run, consecutive slices partition time with no extra stored watermark, each row is classified exactly once, and by the time a row is classified its load state is final — a late-loading fade lands in a future slice and still resets the streak, and a transiently-unfilled order has its fill row by classification time. Went with the trailing-watermark variant over window-derived because it preserves the run-cadence semantics the backtest validated (recovery ≥1h of wall-clock clean activity per level regardless of volume). Tests pin deferred classification, exactly-once classification after the horizon, and the trip path staying on the un-lagged rate window.

lib/cron/fade-rate-v2.ts
codyborn: Fixed in cc7c6d6 exactly as suggested: `newCompletions` now only counts rows with deadline past `fadeWindowStart`, so during-bench fills earn nothing toward the streak whether the cron runs before or after expiry. During-bench *fades* still count in `newFades` (a fade is a fade — it should reset the streak). The in-flight-cohort test now pins both: 3 during-block completions → 0 streak credit, 2 during-block fades → newFades 2.

CLAUDE.md
codyborn: Good catch — fixed the doc to `fillTimeBlocks > 0` plus an explicit keep-in-sync warning (cc7c6d6). Since the extract keeps the raw columns, I re-ran the whole backtest with fades recomputed locally under the correct semantics: 387 rows (1.6%) were misclassified, concentrated in a few fillers (one "26% chronic" was actually a legitimate 8.5% filler). All qualitative conclusions hold — chronic ≥30% faders: 17% benched pre-fix → 59% shipped, allowed fades 605 → 180; collateral on ≤11% fillers ≤0.8% bench — and the corrected numbers replaced the stale ones in the code comments (uncapped worst block 128h not 152h; cap cost ~20% not ~12%). The pre-fix decay-leak stat holds at 57% of decay events granted by fade-containing runs (100% for the extreme faders).

lib/cron/fade-rate-v2.ts
codyborn: Fixed in cc7c6d6: `newConsecutiveBlocks` caps the stored counter at `MAX_BLOCK_BACKOFF_EXPONENT`, so full recovery is bounded at 7 × 6 clean runs (~7h of clean activity) and a reformed filler isn't one sub-threshold fade from a 32h block indefinitely. Kept the `min()` in `calculateBlockUntilTimestamp` as defense for any legacy stored values above the cap. Test pins that blocking at the cap holds both the duration exponent and the counter.

lib/repositories/base.ts
codyborn: Agreed — `consecutiveCleanRuns` is now required in `ToUpdateTimestampRow` (cc7c6d6), with a comment explaining why it must stay that way given full-item puts. The repo test keeps the pre-migration missing-attribute coverage via an explicit `undefined as unknown as number` cast (put omits the attribute → read defaults to 0), so the read-path default stays exercised without weakening the type for real callers.

test/crons/fade-rate-v2.test.ts
codyborn: Done in cc7c6d6: added `fadeStats()` and `cbState()` factories with benign defaults; all ~20 literals now spell out only the fields under test (the file shrank by ~170 lines) and the next field addition touches one place.

gretzke (approved):

PR #492, opened 2026-09-03
refactor(compliance): stale-while-revalidate S3 provider; drop complianceListUrl

alanhwu:
## Why

The filler-compliance feature (per-market-maker swapper exclusion lists from S3) stays, but the provider behind it had three problems on the quote hot path:

1. **Every prod quote did an S3 GetObject.** `configs.length === 0` was used as "never loaded"; prod's config is an empty `[]`, so nothing was ever considered loaded (~95k GetObjects/hour fleet-wide).
2. **Refreshes were `await`ed inside the request.** A slow refresh stalled the quote. The 2026-07 incident (dead upstream, ~15s stalls) came from exactly this shape.
3. **A dead sub-feature, `complianceListUrl`**: a per-config HTTP fetch of extra addresses from a market-maker-hosted URL. No fetch of it ever observably succeeded, it caused the July incident, and no config uses it.

## What changed

- **Loaded-ness is tracked explicitly.** An empty config is a loaded config.
- **Stale-while-revalidate.** The first call per container awaits one bounded S3 fetch (the shared bounded client from `lib/util/config-s3-client.ts`). After that, a request that notices the 5-minute interval has elapsed kicks off a single-flight background refresh and returns the current data immediately; the new index swaps in atomically when the fetch completes. The fetch deliberately stays on the request path rather than container init (Dec-2024 S3 clock-skew issue).
- **Fail-open preserved.** A failed first load serves an empty exclusion set; a failed refresh keeps the last good data. Either way the next attempt waits for the interval, so there is no retry storm.
- **Precomputed lookup.** The endpoint → excluded-swapper index is built once at load time, not per request.
- **`complianceListUrl` deleted entirely**: the field, `FillerComplianceList`, the axios call, and `COMPLIANCE_LIST_TIMEOUT_MS`.
- **Interface reduced** to `ensureLoaded(): Promise<void>` + `isExcluded(endpoint, swapper): boolean`. `getConfigs()` and `getEndpointToExcludedAddrsMap()` are gone; `passFillerCompliance()` in `WebhookQuoter` collapses into `isExcluded`. `isExcluded` is synchronous by design so it cannot add latency; `ensureLoaded` exists only because the first-use fetch has to be awaited somewhere, and it resolves immediately on every call after that.
- **Mock** updated to the new interface. Hard-quote still injects `new MockFillerComplianceConfigurationProvider([])` deliberately.
- **`RFQ_PHASE_COMPLIANCE` emission in `WebhookQuoter` is unchanged**; it now times only `ensureLoaded()`.

### The one deliberate behavior change: case-insensitive address matching

Addresses are lowercased on both sides of the comparison (config entries at index-build time, the request swapper at lookup time). Previously this was an exact string match, which silently failed to exclude a swapper whenever the config's checksum casing differed from the request's. Endpoint matching is unchanged (exact URL).

### Line counts

| File | Before | After |
|---|---|---|
| `lib/providers/compliance/s3.ts` | 105 | 69 |
| `lib/providers/compliance/index.ts` | 16 | 34 |
| `lib/providers/compliance/mock.ts` | 24 | 22 |
| **Provider total** | **145** | **125** |
| `test/providers/compliance/s3.test.ts` | 138 | 127 |
| `lib/quoters/WebhookQuoter.ts` | 489 | 481 |

`index.ts` grows because the shared endpoint → excluded-swapper index builder and lookup (used by both the S3 and mock providers) live there next to the interface. Runtime dependencies of the provider: axios removed.

## Tests

`test/providers/compliance/s3.test.ts` is rewritten around an injected hand-written fake S3 client (optional constructor parameter, defaults to the shared bounded client). No `jest.mock()`, no `S3Client.prototype` patching. Covers:

- empty config does **not** trigger a refetch per request
- no refresh before the interval; interval-elapsed refresh does not block the caller; new data swaps in afterward; concurrent callers share one in-flight fetch
- failed refresh keeps the last good data and retries only at the next interval
- failed first load fails open and retries only at the next interval
- case-insensitive matching in both directions; unknown endpoint → not excluded

## Verification

- `yarn build && yarn test:unit && yarn lint`: 36 suites / 347 tests / 38 snapshots pass; `.snap` files untouched; 0 lint errors, and the repo-wide warning count drops 87 → 85 (the two removed `catch (e: any)`).
- `cdk synth` vs `main` (both sides from clean installs, identical converged `cdk.context.json`, synthesized twice each, compared before committing): 13 of 19 templates byte-identical, including every nested stack. The 6 that differ change **only** the Quote and HardQuote Lambda asset hashes and the resulting `AWS::Lambda::Version` logical IDs / alias references (the two bundles that contain this code). Zero resource additions, removals, or type changes.

## What to watch after deploy

1. **`RFQ_PHASE_COMPLIANCE` p99 → ~0 ms.** It now measures only `ensureLoaded()`, which awaits I/O once per container lifetime. Anything above a few ms outside cold starts means the refresh is somehow back on the request path.
2. **The compliance provider's `Fetched configs` log line** (`quoter: S3FillerComplianceConfigurationProvider`) should fall from once per quote request to once per 5 minutes per container. Its `configsLength` payload is unchanged.

## Out of scope (left as-is per the change boundary; worth a follow-up)

- `lib/handlers/hard-quote/injector.ts` comment still mentions `passFillerCompliance()` and "the outbound compliance-list fetch", both now gone.
- `lib/util/config-s3-client.ts` comment says the compliance provider refreshes "inline on the quote path"; it is now background after first load.

### DIFF
--- a/lib/providers/compliance/index.ts
+++ b/lib/providers/compliance/index.ts
@@ -1,15 +1,33 @@
 export interface FillerComplianceConfiguration {
   endpoints: string[];
   addresses: string[];
-  complianceListUrl?: string;
-}
-export interface FillerComplianceList {
-  addresses: string[];
 }
 
 export interface FillerComplianceConfigurationProvider {
-  getConfigs(): Promise<FillerComplianceConfiguration[]>;
-  getEndpointToExcludedAddrsMap(): Promise<Map<string, Set<string>>>;
+  // The first call per container awaits the load; every later call resolves immediately.
+  ensureLoaded(): Promise<void>;
+  // Synchronous by design: called once per endpoint on the quote path after ensureLoaded().
+  isExcluded(endpoint: string, swapper: string): boolean;
+}
+
+// endpoint -> excluded swappers, lowercased on both sides of the comparison so a
+// checksum-cased config entry still matches a lowercased request swapper (and vice versa).
+export type ExclusionIndex = Map<string, Set<string>>;
+
+export function buildExclusionIndex(configs: FillerComplianceConfiguration[]): ExclusionIndex {
+  const index: ExclusionIndex = new Map();
+  for (const { endpoints, addresses } of configs) {
+    for (const endpoint of endpoints) {
+      const excluded = index.get(endpoint) ?? new Set<string>();
+      addresses.forEach((address) => excluded.add(address.toLowerCase()));
+      index.set(endpoint, excluded);
+    }
+  }
+  return index;
+}
+
+export function isExcludedIn(index: ExclusionIndex, endpoint: string, swapper: string): boolean {
+  return index.get(endpoint)?.has(swapper.toLowerCase()) ?? false;
 }
 
 export * from './mock';
--- a/lib/providers/compliance/mock.ts
+++ b/lib/providers/compliance/mock.ts
@@ -1,24 +1,22 @@
-import { FillerComplianceConfiguration, FillerComplianceConfigurationProvider } from '.';
+import {
+  buildExclusionIndex,
+  FillerComplianceConfiguration,
+  FillerComplianceConfigurationProvider,
+  isExcludedIn,
+} from '.';
 
 export class MockFillerComplianceConfigurationProvider implements FillerComplianceConfigurationProvider {
-  constructor(private configs: FillerComplianceConfiguration[]) {}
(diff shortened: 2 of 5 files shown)


alanhwu:
replaced by #494

PR #501, opened 2026-09-07
feat: attribute filler addresses only on posted exclusive hard quotes; drop the 3-address cap

ericneil-sanc:
## Summary

Record filler-address attribution **only when a hard quote is posted with exclusivity**, remove the 3-address-per-filler cap, and give the `FillerAddress` table a shape that stays bounded without one. Supersedes #498, which patched the same incident inside the old design.

1. **`/quote` no longer touches DynamoDB at all.** `WebhookQuoter` loses its repository dependency and the per-response `addNewAddressToFiller` call. The soft-quote handler is unchanged from main.
2. **The hard-quote handler records the winner at the confirmed post**, and only when the cosigned order actually grants exclusivity (an open order carries no filler address). The breaker scores only V2/V3 orders and every one of those is cosigned here, so this is the one place attribution is needed. The write is made by `recordPostedOrder` from the same record as the PostedOrders row: both run concurrently under the existing 500 ms wall on a bounded DynamoDB client (short ceilings; the attribution write gets one SDK retry inside the wall, the PostedOrders write stays at one attempt), and each failure is counted on its own metric (`FILLER_ADDRESS_RECORD_FAILED` for attribution) before the response returns. A claim refused by the first-writer-wins condition is reported back rather than swallowed: it is logged with the existing owner (`ReturnValuesOnConditionCheckFailure: ALL_OLD`) and counted on `FILLER_ADDRESS_CLAIM_REJECTED`, distinct from a DynamoDB failure. A DynamoDB failure can no longer 5xx a response, and a slow one adds at most the wall.
3. **No per-filler cap.** `MAX_FILLER_ADDRESSES` is gone.
4. **One item per address, with a TTL.** The old design kept each filler's addresses in one item as a string set; with no cap a filler that rotated addresses would grow that item toward the 400 KB limit, and set writes cost the whole item's size. Now each address is its own row (`pk` = checksummed address, `filler`, `expiresAt`) with a 30-day TTL refreshed on each win, written by a single conditional `UpdateItem` (`attribute_not_exists(filler) OR filler = :owner`, so first-writer-wins is preserved) and no read. The cron looks up exactly the addresses in its fade rows (BatchGet, chunked by 100, with `UnprocessedKeys` drained with backoff and logged if any remain) and drops mappings to endpoints no longer in the webhook config, warning with counts when any address resolves to no configured filler. The order-service shadow's scorer extends that map for addresses only its own rows name, so rows Redshift has not loaded yet are scored rather than dropped.

## Why

The only reader of this table is the fade-rate breaker, which resolves a **posted order's** exclusive filler address to a webhook. Recording every filler's address on every `/quote` response was pure amplification (~1.7 GetItems per invocation). The dominant cost was one filler that has 4–5 addresses against a cap of 3: every response carrying address 4 or 5 hit the "cap reached" branch, paid two reads, and could never be registered, so it repeated forever. That throttled the table on 2026-09-07 and, via the un-awaited write, 5xx'd `/quote`.

The cap never protected quoting. It bounded the size of the per-filler set item and hid a filler's activity beyond its first 3 addresses from the breaker.

## Load after this change

- **`/quote`:** zero DynamoDB calls.
- **`/hard-quote`:** one read-free conditional write (at most two attempts) per confirmed exclusive post, concurrent with the PostedOrders write under the same 500 ms wall. At today's order rate that is on the order of 10k writes a day, total.
- **Cron:** one BatchGet per 100 distinct addresses in the 24h fade window, every 10 minutes.

Table capacity (70 RCU / 250 WCU) is untouched here; both are now far above need and can be trimmed in a follow-up.

## Follow-ups (not in this PR)

- **Indeterminate post path.** Attribution (like the PostedOrders row) is written only on a confirmed 200/201. If the order service accepted an order but the post timed out past the 2.5 s reconcile, the winner's address is not recorded until its next confirmed win. Low impact since fillers reuse 1–5 addresses; noted for completeness.
- **One-off backfill.** Rows written before the TTL existed have no `expiresAt`; an address whose owner never wins again keeps its row. A script stamping `expiresAt` on address rows and deleting the legacy URL-keyed set rows would remove both caveats.
- **Endpoint migration.** A filler whose webhook URL changes keeps losing claims on its old addresses to the old URL's rows (first-writer-wins) until those rows expire 30 days after the old URL's last win; `FILLER_ADDRESS_CLAIM_REJECTED` makes this visible, and deleting the rows by hand is the fix until then.
- **PostedOrders already holds the endpoint per order.** The Redshift path is the only reason this table still exists. The shadow breaker's `buildFadeRows` should carry `record.filler` through instead of round-tripping via FillerAddress, and once the PostedOrders breaker is primary this table and its metric can be retired.

## Compatibility / rollout

- **No data migration.** Existing address rows carry entity `addressToFillerEntity` and keep parsing; they have no `expiresAt` and persist until their next win stamps one. Legacy filler → [addresses] rows (keyed by endpoint URL) are never requested by address and are inert; they can be deleted at leisure.
- **CDK:** `timeToLiveAttribute: 'expiresAt'` on `FillerAddrTable`. Enabling TTL on an existing table is an in-place update.
- **Behaviour change for the breaker:** addresses that only ever appeared on losing quotes, or on soft quotes that never became an order, are no longer recorded. Neither can appear on a posted order, so breaker results are unchanged. The backtest scan in `CLAUDE.md` will see fewer addresses for the same reason.
- If #498 has merged first, this branch conflicts in `filler-address-repository.ts` and `WebhookQuoter.ts`; resolution is "take this branch" for both files.

## Test plan

- [x] `test/repositories/filler-address-repository.test.ts` (DynamoDB Local): record/resolve, checksumming, TTL stamp, **12 addresses for one filler**, 105-address batch chunking, first-writer-wins refused-not-thrown and reported with the existing owner, TTL refresh is one write and no read, a refused claim changes neither owner nor TTL, legacy address rows in their real stored shape (`_et`/`_ct`/`_md`, no `expiresAt`) resolve without migration, a BatchGet page with no entry for the table is warned about rather than read as no rows.
- [x] `test/handlers/hard-quote/posted-order-recorder.test.ts`: attribution written alongside the record; none for open/non-exclusive orders; a failing attribution write is counted on its own metric and leaves the record intact; a refused claim is counted on `FILLER_ADDRESS_CLAIM_REJECTED`, not as a write failure; a hung attribution write is bounded by the wall; a failing record write does not prevent attribution.
- [x] `test/handlers/hard-quote/handler.test.ts`: attribution on a confirmed exclusive post; none when the order posts open; a rejecting repository still yields 200 (no `setImmediate` needed: the write is awaited).
- [x] `test/repositories/filler-address-repository.test.ts`: a BatchGet that returns `UnprocessedKeys` is retried and the map is complete.
- [x] `test/handlers/shared/quote-injector.test.ts`: repository present on the hard-quote container, absent from the soft-quote container and from the quoter.
- [x] `test/crons/fade-rate-cron-run.test.ts`: cron looks up exactly the row addresses; a mapping to a removed endpoint benches nobody; the shadow's scorer resolves an address only its rows name and blocks that filler, requesting exactly the missing address.
- [x] `test/handlers/quote/*`: unchanged from main apart from dropping the quoter's constructor argument.
- [x] Full unit suite green; `tsc --noEmit`, prettier, eslint clean.
- [ ] After deploy: `FillerAddress` `ConsumedReadCapacityUnits` ≈ 0 outside cron ticks, `ReadThrottleEvents` = 0, `FILLER_ADDRESS_RECORD_FAILED` = 0, `FILLER_ADDRESS_CLAIM_REJECTED` = 0 (a sustained non-zero rate names a filler on another's address or a changed webhook URL), item count plateaus at (distinct winning addresses in 30d).


lib/cron/fade-rate-v2.ts
```diff
@@ -249,7 +266,7 @@ export async function runFadeRateCron(metrics: MetricsLogger, deps: FadeRateCron
 
   if (result) {
     const fillerEndpoints = webhookProvider.fillerEndpoints();
-    const addressToFillerMap = await fillerAddressRepo.getAddressToFillerMap(fillerEndpoints);
+    const addressToFillerMap = await lookupFillersForRows(result, fillerEndpoints, fillerAddressRepo);
```
codyborn: **Shadow evaluator is blinded to rows only the new source has.** The address→filler map is now built from the Redshift rows' addresses only, but the `score` closure below captures the same map and scores order-service (PostedOrders) rows with it. An address present in PostedOrders but absent from Redshift (batch-load lag, which `STREAK_FINALITY_LAG_SECS` already compensates for, or `FADE_QUERY_ROW_LIMIT` truncation) never makes it into the lookup, `getFillersFadeStats` hits `if (!fillerHash) return` with no `log` passed, and the shadow's `decisionsVsRestricted`/`wouldBlock` agree with production by construction. On main `getAddressToFillerMap(fillerEndpoints)` returned every configured filler's addresses regardless of Redshift.

The `ShadowContext` doc comment ("shadow decisions differ from the real ones only through the rows") is no longer true, and no cron test asserts shadow behaviour.

Fix: build the shadow's map from the union of Redshift and order-service row addresses, or carry `record.filler` through `buildFadeRows` so the shadow never consults FillerAddress (which the PR already lists as a follow-up).

lib/repositories/filler-address-repository.ts
```diff
@@ -4,225 +4,223 @@ import { Entity, Table } from 'dynamodb-toolbox';
...
+            { or: true, attr: 'filler', eq: filler },
+          ],
+          execute: true,
+        }
+      );
+    } catch (err) {
+      if ((err as Error)?.name !== 'ConditionalCheckFailedException') {
+        throw err;
+      }
+      DynamoFillerAddressRepository.log.info(
```
codyborn, line 150: **A rejected claim is classified as a fulfilled write, with no signal and no owner in the log.** When a filler keeps address X but its configured `endpoint` changes (configSignature in s3.ts tracks `endpoint`, and FillerCBTimestampsV2 is keyed by that same string), the row `{pk: X, filler: URL_old}` fails the `attribute_not_exists(filler) OR filler = :owner` condition on every win from URL_new. That is caught here, logged at info with only `attemptedOwner` (main logged `existingOwner`; UpdateItem can't know it without `returnValues: 'all_old'`), and resolves, so `attributed.status === 'fulfilled'` in posted-order-recorder.ts and `FILLER_ADDRESS_RECORD_FAILED` never fires. The cron then maps X→URL_old, `configured.has(URL_old)` is false, and X's fades bench nobody, again with no warn or counter.

Also: the row's `expiresAt` is never refreshed since URL_old never wins again, which contradicts the api-stack comment "rows expire … after the filler's last win with them", and legacy rows with no `expiresAt` freeze forever. Main's sentence "A genuine address migration between endpoints needs a manual override" was deleted without a replacement path.

Suggest: request `ALL_OLD` and log the existing owner, emit a counter with a reason distinct from write failure on the rejected claim and on each `configured`-filter drop, or switch to last-writer-wins now that the source is the trusted cosigned order.

test/repositories/filler-address-repository.test.ts
```diff
@@ -22,126 +26,178 @@ const documentClient = DynamoDBDocumentClient.from(new DynamoDBClient(dynamoConf
...
-    // MAX_FILLER_ADDRESSES + 1 distinct addresses, disjoint from the fixtures above
-    const addrs = Array.from({ length: MAX_FILLER_ADDRESSES + 1 }, (_, i) =>
-      getAddress(`0x${(0x1000 + i).toString(16).padStart(40, '0')}`)
+  it('still resolves address rows written before per-address TTLs existed (no migration)', async () => {
+    // Shape of a pre-TTL row: dynamodb-toolbox entity marker, no expiresAt.
+    const legacy = addr(0x4001);
+    await documentClient.send(
+      new PutCommand({
+        TableName: 'FillerAddress',
+        Item: { pk: legacy, filler: FILLER2, entity: 'addressToFillerEntity' },
```
codyborn: **This fixture doesn't have the shape of a real legacy row, so it exercises the wrong branch.** dynamodb-toolbox stores the entity marker as `_et` (plus `_ct`/`_md`); `entity` is an output-only alias. Main's `addressToFillerEntity.put({pk, filler})` produced `{pk, filler, _et: 'addressToFillerEntity', _ct, _md}`. In `Table.parseBatchGetResponse`, `Tables[table][item['_et']]` is undefined for this fixture so it takes the `else { return item }` raw passthrough; a real row goes through `addressEntity.parse()`.

Prod reads work today, but the PR removed every test that wrote through the real entity and nothing remaining reads back a row of the true legacy shape (no `expiresAt`, `_et` present). A future schema change that makes `parse` drop or reject `filler`/`expiresAt` on legacy rows keeps this suite green while the cron gets an empty map. The test-plan claim "legacy address rows resolve without migration" isn't proven by this.

Seed `{pk, filler, _et: 'addressToFillerEntity', _ct, _md}` (or put through main's entity definition). Separately, the legacy set-row fixture at L192 is keyed by FILLER1, which the interference test never requests, so it can't interfere.

lib/repositories/filler-address-repository.ts
```diff
@@ -4,225 +4,223 @@ import { Entity, Table } from 'dynamodb-toolbox';
...
-    }
-
-    // Already registered to this owner — idempotent no-op.
-    if (existingOwner === owner) {
-      return;
+  async getAddressToFillerMap(addresses: string[]): Promise<Map<string, string>> {
+    const unique = [...new Set(addresses.map((a) => getAddress(a)))];
+    const resMap = new Map<string, string>();
+    const collect = (page: BatchGetPage) =>
+      ((page.Responses?.FillerAddress ?? []) as AddressItem[]).forEach((row) => {
```
codyborn: **Hardcoded `Responses.FillerAddress` plus `?? []` fails open.** dynamodb-toolbox keys `Responses` by the Table's physical name, which here comes from `DYNAMO_TABLE_NAME.FILLER_ADDRESS`. If that constant is ever suffixed or renamed, BatchGet succeeds, this reads `undefined`, the map is empty, "resolved: 0" is logged at info, and every filler is unscored for the run. Main's `items.FillerAddress.forEach` threw in that case.

Use `page.Responses?.[DYNAMO_TABLE_NAME.FILLER_ADDRESS]` and warn when the key is absent while `unique.length > 0`.

lib/handlers/hard-quote/injector.ts
```diff
@@ -37,8 +42,10 @@ export class QuoteInjector extends ApiInjector<ContainerInjected, RequestInjecte
     return {
       ...base,
       orderServiceProvider: new UniswapXServiceProvider(log, orderServiceUrl),
-      // Builds its own bounded DynamoDB client; construction is lazy (no I/O).
+      // Both build their own bounded DynamoDB client (the writes sit in series with the
+      // response); construction is lazy (no I/O).
       postedOrderRepository: DynamoPostedOrderRepository.create(),
+      fillerAddressRepository: DynamoFillerAddressRepository.create(),
```
codyborn, line 48: **Attribution durability went from 3 SDK attempts on every RFQ response to 1 attempt, 300ms ceiling, only at a confirmed 200/201.** `create()` defaults to `postedOrderDocumentClient()` (maxAttempts 1) and is only invoked inside the 200/201 branch of handler.ts; `uniswapxService.ts` returns 500 on a timed-out post whose reconcile GET misses, so an accepted order can get no attribution. A throttle at the moment of a new address's first win leaves that address unmapped ("filler address not found dynamo mapping", info only) until it wins again.

Impact is modest for 1–5-address fillers (a stable address already has a row; a miss just skips a TTL refresh), and the indeterminate-post gap is disclosed in the description, but the 3→1 attempt reduction isn't called out as a trade-off. Consider a second attempt within the 500ms wall, or see the comment on `lookupFillersForRows` about using PostedOrders as the source of truth instead.

lib/cron/fade-rate-v2.ts
```diff
@@ -222,6 +222,23 @@ function buildOrderServiceShadow(metrics: MetricsLogger): FadeRateCronDeps['shad
   return (ctx) => runFadeRateShadow(ctx, { source, log: shadowLog, metrics });
 }
 
+/**
+ * Resolves the filler (webhook) behind every address in the fade rows. Looks up exactly the
+ * addresses the rows mention (one BatchGet per 100), then keeps only mappings to endpoints in
+ * the current webhook config: an address whose filler has since been removed from the config
+ * must not resurrect a timestamp row for it.
+ */
+export async function lookupFillersForRows(
```
codyborn, line 233: **Is this whole mechanism needed, given PostedOrders already holds the same fact?** The PR adds a TTL attribute, a conditional UpdateItem on the request path, a second bounded client, BatchGet chunking with an UnprocessedKeys drain, the `configured` post-filter, a new failure metric and three deferred follow-ups, for a table the description says exists "only until the Redshift path is retired".

But `recordPostedOrder` already writes the identical (fillerAddress, filler) pair to PostedOrders from the same `buildPostedOrderRecord` call ("Both rows state the same fact"), `PostedOrderRecord.filler` is documented as the same key space as `FillerAddress.filler`, `getFillerOrdersByDeadline` on the filler-deadline GSI exists, `OrderServiceFadesSource.loadCompletedWindow` already iterates `fillerEndpoints()` over the last 24h, `POSTED_ORDER_TTL_SECS` (48h past deadline) covers the Redshift view's 24h lookback, and `buildOrderServiceShadow` already constructs a PostedOrders repository in this cron.

`new Map(records.map(r => [r.fillerAddress, r.filler]))` over that query would replace this function, `recordWinningAddress`, the configured filter (inherent), the drain machinery, `FILLER_ADDRESS_TTL_SECS` and the CDK TTL change, the injector wiring and `FILLER_ADDRESS_RECORD_FAILED`, and it fixes the shadow-blinding and endpoint-change issues at the root. The one residual consumer is the 28-day backtest extract in CLAUDE.md (`aws dynamodb scan`), which already contradicts the repository doc's "exactly one reader" and would need longer PostedOrders retention or a single cheap write kept for it.

Not blocking on this, but worth a deliberate decision before the follow-ups land.

CLAUDE.md
```diff
@@ -62,8 +62,8 @@ full harness design, per-filler duty-cycle/allowed-fades metrics, and baseline n
 
 Extract query (matches the breaker's fade semantics from `V2_FADE_RATE_SQL`, but with **no
 24h window, no latest-100 cap, and no row limit** — the replay applies windowing itself).
-**Keep the `faded` CASE in sync with `V2_FADE_RATE_SQL`** — e.g. #461 changed Dutch_V3 to
-`fillTimeBlocks > 0` (a fill at the decay-start block is _not_ a fade); an extract using the
+**Keep the `faded` CASE in sync with `V2_FADE_RATE_SQL`** — e.g. #461 changed Dutch*V3 to
+`fillTimeBlocks > 0` (a fill at the decay-start block is \_not* a fade); an extract using the
```
codyborn: **Autoformat churn corrupted the ordertype literal.** Main reads `changed Dutch_V3 to \`fillTimeBlocks > 0\` (a fill at the decay-start block is _not_ a fade)`. Under CommonMark this now renders as "Dutch<em>V3 … is _not</em> a fade": the identifier the SQL actually uses (`ordertype = 'Dutch_V3'`, fades-repository.ts) no longer appears in the paragraph telling readers to keep it in sync, and a literal `\_` shows. Prettier 2.8.8 over main's file reproduces this byte-for-byte, and `lint:prettier` only globs `**/*.ts`, so no gate requires it.

Please revert these two lines and keep only the intended FillerAddress paragraph edit.

lib/handlers/hard-quote/posted-order-recorder.ts
```diff
@@ -104,24 +111,54 @@ export async function recordPostedOrder(args: RecordPostedOrderArgs): Promise<vo
...
   }
 }
 
+const errorMessage = (e: unknown): unknown => (e instanceof Error ? e.message : e);
+
 // Promise.race subscribes to the losing promise too, so a write that fails after the
 // timeout fired is swallowed rather than surfacing as an unhandled rejection. The timer is
 // always cleared so a fast write leaves nothing pending on the event loop.
-async function withTimeout<T>(promise: Promise<T>, ms: number): Promise<T> {
+async function withTimeout<T>(promise: Promise<T>, ms: number, label: string): Promise<T> {
```
codyborn: **Duplicate of the private `withTimeout` in `lib/cron/fade-rate-shadow.ts`** (same Promise.race, same cleared timer, same unhandled-rejection comment), and this PR changes only this copy's signature by adding `label`. The next fix (a typed `TimeoutError` so callers can tag a timeout reason, or an `AbortSignal` so the losing DynamoDB call actually stops) will land in one and not the other.

Hoist a single `withTimeout(promise, ms, label)` into `lib/util/time.ts` next to the `sleep` this PR already imports from there, and use it in both call sites.

lib/cron/fade-rate-v2.ts
```diff
@@ -249,7 +266,7 @@ export async function runFadeRateCron(metrics: MetricsLogger, deps: FadeRateCron
 
   if (result) {
     const fillerEndpoints = webhookProvider.fillerEndpoints();
-    const addressToFillerMap = await fillerAddressRepo.getAddressToFillerMap(fillerEndpoints);
+    const addressToFillerMap = await lookupFillersForRows(result, fillerEndpoints, fillerAddressRepo);
     const fillerTimestamps = await timestampDB.getFillerTimestampsMap(fillerEndpoints);
```
codyborn: **Serial awaits on independent work.** `lookupFillersForRows` (a BatchGet plus up to 350ms of UnprocessedKeys backoff per chunk) and `getFillerTimestampsMap` both depend only on `result` and `fillerEndpoints`, which are already in hand. Every run pays an extra full DynamoDB round trip of wall time for nothing.

```ts
const [addressToFillerMap, fillerTimestamps] = await Promise.all([
  lookupFillersForRows(result, fillerEndpoints, fillerAddressRepo),
  timestampDB.getFillerTimestampsMap(fillerEndpoints),
]);
```

lib/repositories/filler-address-repository.ts
```diff
@@ -4,225 +4,223 @@ import { Entity, Table } from 'dynamodb-toolbox';
...
+  /**
+   * Attribute `address` to `filler` because a quote from `filler` carrying `address` won.
+   * First-writer-wins: an address already attributed to a different filler keeps its owner (a
+   * claim from another endpoint is logged and ignored, so one filler cannot poison another's
+   * attribution). Re-recording for the same owner refreshes the TTL.
+   */
+  recordWinningAddress(address: string, filler: string): Promise<void>;
+  /** address -> filler for exactly the addresses asked (keys checksummed); unknown addresses absent. */
+  getAddressToFillerMap(addresses: string[]): Promise<Map<string, string>>;
+  getFillerByAddress(address: string): Promise<string | undefined>;
```
codyborn: **No production caller.** `getFillerByAddress` is kept on the rewritten interface and implemented twice (Dynamo and Mock), but only the repository test uses it, while the interface's own docblock says the table has exactly one reader via the batch map. Drop it from the interface and both implementations; the test can go through `getAddressToFillerMap([addr])` (which it already does in most cases) or the raw-item helper.

lib/quoters/WebhookQuoter.ts
```diff
@@ -344,10 +342,12 @@ export class WebhookQuoter implements Quoter {
...
 
-      // do not await to minimize latency
-      if (response.filler) {
-        this.repository.addNewAddressToFiller(response.filler, endpoint);
-      }
+      // Filler-address attribution is NOT done here: only a winning quote that is then posted
+      // with exclusivity puts a filler address on an order, so recordPostedOrder writes it at
+      // the confirmed post (see handlers/hard-quote/posted-order-recorder.ts). Doing it per
+      // response cost a DynamoDB round trip for every filler on every quote and throttled the
+      // table on 2026-09-07.
```
codyborn: nit: drop the point-in-time reference. "throttled the table on 2026-09-07" is provenance and belongs in the commit message / PR description; the comment should describe current behaviour only (e.g. "…cost a DynamoDB round trip for every filler on every quote and throttles the table under load").

bin/stacks/api-stack.ts
```diff
@@ -440,6 +440,11 @@ export class APIStack extends cdk.Stack {
       deletionProtection: true,
       pointInTimeRecovery: true,
       contributorInsightsEnabled: true,
+      // Address rows expire FILLER_ADDRESS_TTL_SECS after the filler's last win with them, so
+      // rows written from now on cannot grow the table without bound. Rows written before the
+      // attribute existed have no expiresAt: they get one on their next win, and an address
```
codyborn: nit: same point-in-time phrasing here. "rows written from now on" and "Rows written before the attribute existed" describe a rollout moment rather than the current invariant. Something like: "Rows carry `expiresAt` and expire FILLER_ADDRESS_TTL_SECS after the filler's last win with them. A row without `expiresAt` gets one on its next win; an address that never wins again keeps its row until a one-off backfill stamps it." Keeps the same content, reads the same in a year.

codyborn (commented):
Automated review (Claude) of the final diff at 92f760d. Ten findings plus two nits, inline, most severe first:

1. Shadow evaluator is blinded to PostedOrders-only addresses (`fade-rate-v2.ts` L269).
2. Endpoint-change claim rejections are swallowed as fulfilled writes with no owner logged and no metric (`filler-address-repository.ts` L126).
3. Legacy-row test fixture has the wrong shape and exercises the raw-passthrough branch (`filler-address-repository.test.ts` L145).
4. Hardcoded `Responses.FillerAddress ?? []` fails open on a table-name change (L142).
5. Attribution durability regressed from 3 attempts to 1, not called out (`injector.ts` L48).
6. PostedOrders already holds the same (address, filler) fact; the added FillerAddress mechanism may be avoidable (`fade-rate-v2.ts` L231).
7. Prettier churn in CLAUDE.md corrupted `Dutch_V3` → `Dutch*V3`; revert L65-66.
8. `withTimeout` duplicated with `fade-rate-shadow.ts`; hoist to `lib/util/time.ts`.
9. Two independent awaits run serially in the cron (L269-270).
10. `getFillerByAddress` has no production caller.

Nits: two point-in-time comments ("on 2026-09-07", "from now on") should describe current behaviour only.

Considered and dropped: the IAM-grant gap (pre-existing, unchanged), a synchronous throw inside `allSettled` (every implementation is `async`), and a two-clients cold-start concern (both writes are equally cold).

lib/cron/fade-rate-v2.ts
codyborn: Addressed in 07e46f0. `ShadowContext.score` is now async and extends the real run's map for any address its rows name that the Redshift rows did not (`extendFillerMapForRows`, one extra BatchGet only for the missing addresses). New cron test: a filler whose orders exist only in the shadow's rows gets a block decision, and the lookup log shows exactly `[[A, B], [C]]`.

lib/repositories/filler-address-repository.ts
codyborn: Addressed in 07e46f0. `recordWinningAddress` returns `{ outcome: 'recorded' } | { outcome: 'owned_by_other', existingOwner }`; the update passes `ReturnValuesOnConditionCheckFailure: 'ALL_OLD'` so the refusal names the owner without a second round trip (verified against DynamoDB Local). The recorder counts a refusal on a new `FILLER_ADDRESS_CLAIM_REJECTED` metric with a warn log, distinct from `FILLER_ADDRESS_RECORD_FAILED`. The cron warns with counts when addresses resolve to no configured filler. Interface doc restores the manual-delete migration path and the api-stack TTL comment now states that a refused claim does not refresh the row. Kept first-writer-wins rather than last-writer-wins so a filler still cannot poison another's attribution; the PR description has a new follow-up bullet on endpoint migration.

test/repositories/filler-address-repository.test.ts
codyborn: Addressed in 07e46f0. The fixture is now `{ pk, filler, _et: 'addressToFillerEntity', _ct, _md }`, the exact shape toolbox's `put()` wrote on main, so the row takes the `Entity.parse()` branch. The set-row interference test was removed: `getAddressToFillerMap` checksum-validates its keys, so a URL-keyed row can never be requested and the test could not fail.

lib/repositories/filler-address-repository.ts
codyborn: Addressed in 07e46f0. Reads `page.Responses?.[DYNAMO_TABLE_NAME.FILLER_ADDRESS]`; a page with no entry for the table is counted and warned about (`pagesWithoutTable`) instead of being read as zero rows. New DynamoDB Local test mocks a page keyed by another table name and asserts the warn.

lib/handlers/hard-quote/injector.ts
codyborn: Addressed in 07e46f0. `postedOrderDocumentClient` takes `maxAttempts`; the FillerAddress repository defaults to `FILLER_ADDRESS_WRITE_MAX_ATTEMPTS = 2` under the same 200/300 ms per-request ceilings, so a throttled first attempt gets one retry inside the recorder's 500 ms wall. The PostedOrders write stays at one attempt as before. The description's summary and load sections now state this.

lib/cron/fade-rate-v2.ts
codyborn: Deferred, per discussion with the reviewer. This PR keeps the FillerAddress mechanism and hardens it (see the other threads); the PostedOrders consolidation stays as the listed follow-up, where the shadow's `buildFadeRows` carries `record.filler` through and this table is retired with the Redshift path. The 28-day backtest extract is the one reader that needs a home before that.

CLAUDE.md
codyborn: Addressed in 07e46f0. Lines 65-66 restored to main's text (`Dutch_V3`, `_not_`); only the intended FillerAddress paragraph edit remains in the file.

lib/handlers/hard-quote/posted-order-recorder.ts
codyborn: Addressed in 07e46f0. One `withTimeout(promise, ms, label)` in `lib/util/time.ts` next to `sleep`; both private copies are gone and the shadow passes `'fade shadow'` as its label.

lib/cron/fade-rate-v2.ts
codyborn: Addressed in 07e46f0. `lookupFillersForRows` and `getFillerTimestampsMap` now run under one `Promise.all`.

lib/repositories/filler-address-repository.ts
codyborn: Addressed in 07e46f0. `getFillerByAddress` removed from the interface and both implementations; the repository test resolves single addresses through `getAddressToFillerMap([addr])`.

lib/quoters/WebhookQuoter.ts
codyborn: Addressed in 07e46f0. Reworded to describe the current cost without the date.

bin/stacks/api-stack.ts
codyborn: Addressed in 07e46f0. Reworded to state the invariant (rows carry `expiresAt`, refreshed on the owner's win, not by a refused claim) without the rollout framing.

codyborn (approved):

PR #503, opened 2026-09-08
fix(fades): paginate PostedOrders index queries; fills without timing carry no verdict

alanhwu:
Superseded by #504, which carries this commit (rebased onto main) as its first commit. Description removed; see #504.


alanhwu:
Superseded by #504 (same fix as its commit 1, plus the promotion).

PR #505, opened 2026-09-08
refactor: explicit per-request observability Context for the quote handler layer (1/3)

alanhwu:
## Summary

Step 1 of 3 in replacing the smart-order-router observability module globals (`setGlobalLogger` / `setGlobalMetric`) with an explicit per-request context. Those globals hold request-scoped state in module scope, which is safe only because Lambda serves one request per container; the ECS target is concurrent, so they have to go. This PR lays the seam and switches **only the handler layer** onto it. Nothing below the handlers changes, the globals are still set, and the wire behaviour (response bodies, log records, metric names/units/dimension sets) is unchanged.

## What's new: `lib/observability/`

Shaped as the subset of the monorepo's interfaces this service uses (backend `packages/lib/uni/interface.ts`), so the port swaps the types, not the call sites:

- `Logger` — the `ILogger` subset: message-first `debug` … `fatal(msg, ...extra)`, `child()`, `setDefaultExtra(...extra)`. bunyan stays the implementation behind `BunyanLogger`, a small adapter that flips `(msg, ...extra)` into bunyan's `(fields, msg)`; an `Error` extra lands under `err` for the standard serializer.
- `Metrics` — the `IMetrics` subset: `count(name, val?, opts?)`, `timer(name, val, opts?)`, `gauge(name, val, opts?)`, all `Promise<void>`, with the monorepo's `MetricOptions` (`tags: string[]`, `dimensions`, `sampleRate`).
- `Context` — `{ logger, metrics, requestId }`, the monorepo `Context`'s porcelain.
- `EmfMetrics` — `Metrics` over the request's existing aws-embedded-metrics `MetricsLogger`.

### `EmfMetrics` delegation design

- **Delegation, not a new sink.** It wraps the very `MetricsLogger` the injector has already configured (namespace + the `Service` / `Service`+`ChainId` / dimensionless dimension sets), so a handler metric lands in the same EMF blob, under the same dimension sets, as before. Dimension sets are not touched in this PR.
- **Same tuples.** `count → (name, val=1, Count)`, `timer → (name, ms, Milliseconds)`, `gauge → (name, val, None)` — the same strings `MetricLoggerUnit` produced. The unit test asserts this against the old `AWSMetricsLogger.putMetric` path on a real `MetricsLogger`, for every handler-level metric.
- **`opts` accepted and dropped.** EMF has no per-metric tags: dimension sets are the injector's, and EMF properties are per-blob (two metrics in one request with different `status:` tags would overwrite each other). The StatsD sink applies them.
- **Never throws, never rejects, emits synchronously.** `putMetric` validates (non-finite value, bad name/unit) and before this PR that exception surfaced as a 500 on the quote; `EmfMetrics` logs a `warn` and drops the metric. Emission happens synchronously inside the async methods, so a `void`-ed call in a `finally` has landed before the caller returns (tested).

### Monorepo alignment (per review)

- Method shapes match `ILogger` / `IMetrics`; an implementation of either is assignable to the local type.
- `timer`, not `hist`, for latencies: the monorepo's rule is `timer()` for durations, services call `timer` ~20× more often than `hist`, and `hist` throws "Method not implemented" in both Lambda `IMetrics` implementations while `CloudWatchMetric.timer` maps to Milliseconds — this repo's mapping.
- Call sites `await` in the main flow and `void` in `catch`/`finally`, matching the monorepo's `no-floating-promises` lint and its fire-and-forget rule for metrics in failure paths.
- Tags use the monorepo's string encoding (`status:failure`, `reason:<enum>`); the monorepo requires them on every operation, and they slot in when PR 2 moves the failure paths. The earlier typed `MetricTags` / `MetricReason` are gone (nothing used them).
- Still changes at port time regardless of interface shape: metric names (the monorepo prefixes the service and uses dotted names; every distribution needs an allowlist + Pulumi tag configuration), the test framework (vitest, `buildTestContext()`), and the `requestBody` log binding.

## What moved to `ctx`

- `lib/handlers/shared/quote-injector.ts` builds `ctx` in `buildQuoteRequestInjected`, alongside the existing global setup, for both quote Lambdas: `ctx.logger` is `BunyanLogger` over the same bunyan child (same `requestId` / `requestBody` bindings), `ctx.metrics` is `EmfMetrics` over the same `MetricsLogger`. `RequestInjected` gains `ctx`.
- `lib/handlers/quote/handler.ts`, `lib/handlers/hard-quote/handler.ts`, `lib/handlers/hard-quote/posted-order-recorder.ts` emit metrics via `ctx.metrics` and log via `ctx.logger`; their `@uniswap/smart-order-router` type imports (`IMetric`, `MetricLoggerUnit`) are gone. `recordPostedOrder` takes `ctx` in place of `log` + `metric`.
- **Log records are byte-identical.** Message-first calls write the same bunyan record as the fields-first calls they replace. The three analytics event lines (`QuoteRequest`, `HardRequest`, the open-order `HardResponse`) pass `''` as the message: bunyan writes `"msg":""` for a fields-only call, so the record is unchanged (asserted in `bunyan-logger.test.ts`), and the CloudWatch subscription filters that feed Redshift key on `$.eventType`, not the message.

## What deliberately still reads the global, and which PR moves it

- `WebhookQuoter` and `S3WebhookConfigurationProvider` read the smart-order-router module globals (`metric`, `log`) → **PR 2 (quote path)**. `setGlobalLogger` / `setGlobalMetric` therefore still run in the injector.
- `getBestQuote` takes the bunyan `Logger` and `IMetric` positionally, so the handlers pass `requestInjected.log` / `metric` through to it (the same objects `ctx` wraps); `RequestInjected.metric` stays for that reason → **PR 2**.
- `lib/cron/fade-rate-v2.ts` binds the global so the config provider's `RFQ_CONFIG_CHANGED` has somewhere to go → **PR 3 (cron)**.
- `APIGLambdaHandler` (base) puts `HANDLER_DURATION` / validation-path `QUOTE_400`/`QUOTE_500` on the raw `MetricsLogger`; unchanged here.
- `@uniswap/smart-order-router` stays in `package.json`.

## Tests

- `test/fakes/`: recording `FakeMetrics` and `FakeLogger` (both `implements` the interface, per the monorepo's test-double rule), plus a `fakeContext()` helper; used in every test touched, no new `jest.mock()`.
- `test/observability/emf-metrics.test.ts`: for each of the 14 handler-level metrics, the new call reaches a real `MetricsLogger` as the identical `(name, value, unit)` the `IMetric` path produced; `count` defaults to 1; `gauge` → `None`; `opts` change neither the tuple nor properties/dimension sets; synchronous emission; a rejected metric is logged and dropped rather than thrown or rejected.
- `test/observability/bunyan-logger.test.ts`: a message-first call writes the same record as bunyan's fields-first call; an empty message reproduces a fields-only record byte for byte; `Error` extras land under `err`; `child()` inherits bindings and `setDefaultExtra` stays on the logger it was set on, leaving the wrapped bunyan logger untouched.
- Handler tests assert the exact metric sequence per path through `FakeMetrics` (200; 404 / no-quote; order-service 4xx including the insufficient-funds carve-out; 5xx; exclusive post with the recorder's bookkeeping metrics landing before `QUOTE_LATENCY`) and the request / error log lines through `FakeLogger` with unchanged bindings. The injector test asserts `ctx.logger` writes through the same child logger and `ctx.metrics` forwards to the request's `MetricsLogger`.
- Contract snapshots (`test/handlers/*/response-surface.test.ts`) pass with the `.snap` files unmodified.

## Verification

- `yarn build && yarn test:unit && yarn lint`: 43 suites / 474 tests / 38 snapshots pass; lint 0 errors and the same 84 pre-existing warnings as `main`, identical per file (none added). No `@ts-ignore`, `eslint-disable` or `any` added.
- `cdk synth` (two worktrees, clean `--frozen-lockfile` installs, identical converged context, two in-tree synths per side), first against `main` and then against the previous head of this PR: **byte-identical templates except the API stack (root + beta + prod assemblies)**, which differs only in the Quote / HardQuote code asset keys, their `CurrentVersion` logical IDs and the `LiveAlias` version pointers — the normal code deploy — plus the pipeline's two asset-publish BuildSpecs (they embed the hashes). Every nested stack (Cron, Analytics, Firehose, Cosigner key, Dashboard) is byte-identical; no resource added, removed or retyped; no alarm / dashboard / dimension change. Bundle module lists: each quote bundle gains exactly `lib/observability/emf-metrics.ts` and `lib/observability/bunyan-logger.ts` (the other new files are type-only); the only content-changed sources are the four handler-layer files; no `node_modules` change.

## Post-deploy signals

- Handler metrics still present in CloudWatch under their existing dimension sets (`Service=SoftQuote|HardQuote`, `Service`+`ChainId`, and the dimensionless rollup): `QUOTE_REQUESTED`, `QUOTE_200/400/404/500`, `QUOTE_POST_ATTEMPT`, `QUOTE_POST_ERROR`, `QUOTE_LATENCY`, `QUOTE_E2E_LATENCY`, `POSTED_ORDER_RECORDED` / `RECORD_FAILED` / `RECORD_LATENCY`, `FILLER_ADDRESS_RECORD_FAILED` / `CLAIM_REJECTED`.
- The quote-latency SEV3 alarm still evaluating (not `INSUFFICIENT_DATA`).
- Zero Lambda errors on both quote functions; `QUOTE_LATENCY` p99 unchanged.
- Redshift `rfqrequests` / `hardrequests` / `hardresponses` row counts unchanged (the analytics event lines are byte-identical, but this is the end-to-end check).
- No `Dropped metric rejected by aws-embedded-metrics` warn lines in the quote Lambda logs (that path was a 500 before this PR, so any occurrence is worth a look).

### DIFF
--- a/lib/handlers/hard-quote/handler.ts
+++ b/lib/handlers/hard-quote/handler.ts
@@ -1,7 +1,6 @@
 import { KMSClient } from '@aws-sdk/client-kms';
 import { TradeType } from '@uniswap/sdk-core';
 import { KmsSigner } from '@uniswap/signer';
-import { MetricLoggerUnit } from '@uniswap/smart-order-router';
 import {
   CosignedV2DutchOrder,
   CosignedV3DutchOrder,
@@ -49,7 +48,7 @@ export class QuoteHandler extends APIGLambdaHandler<
     params: APIHandleRequestParams<ContainerInjected, RequestInjected, HardQuoteRequestBody, void>
   ): Promise<ErrorResponse | Response<HardQuoteResponseData>> {
     const {
-      requestInjected: { log, metric },
+      requestInjected: { ctx, log, metric },
       containerInjected: {
         quoters,
         orderServiceProvider,
@@ -59,9 +58,10 @@ export class QuoteHandler extends APIGLambdaHandler<
       },
       requestBody,
     } = params;
+    const { logger, metrics } = ctx;
     const start = Date.now();
 
-    metric.putMetric(Metric.QUOTE_REQUESTED, 1, MetricLoggerUnit.Count);
+    metrics.increment(Metric.QUOTE_REQUESTED);
 
     // QUOTE_LATENCY below fires only on successful order posts (and is alarmed on). The
     // finally makes this metric cover every exit path — no-quote throws, cosigner
@@ -84,12 +84,12 @@ export class QuoteHandler extends APIGLambdaHandler<
 
       // we dont have access to the cosigner key, throw
       if (request.order.info.cosigner !== cosignerAddress) {
-        log.error({ cosignerInReq: request.order.info.cosigner, expected: cosignerAddress }, 'Unknown cosigner');
+        logger.error({ cosignerInReq: request.order.info.cosigner, expected: cosignerAddress }, 'Unknown cosigner');
         throw new UnknownOrderCosignerError();
       }
       // Instead of decoding the order, we rely on frontend passing in the requestId
       //   from indicative quote
-      log.info({
+      logger.info({
         eventType: 'HardRequest',
         body: {
           requestId: request.requestId,
@@ -110,6 +110,9 @@ export class QuoteHandler extends APIGLambdaHandler<
 
       let bestQuote;
       if (!requestBody.forceOpenOrder) {
+        // The quote path below the handler still takes the bunyan logger and the
+        // smart-order-router IMetric positionally (moved to ctx in a later PR); both are the
+        // same objects ctx wraps, so its logs and metrics are unchanged.
         const result = await getBestQuote(quoters, request.toQuoteRequest(), log, metric, provider, RESPONSE_LOG_TYPE);
         bestQuote = result.bestQuote;
         if (!bestQuote && !requestBody.allowNoQuote) {
@@ -122,17 +125,17 @@ export class QuoteHandler extends APIGLambdaHandler<
       let cosignerData: CosignerData | V3CosignerData;
       if (bestQuote) {
         cosignerData = await getCosignerData(request, bestQuote, orderType, provider);
-        log.info({ bestQuote: bestQuote }, 'bestQuote');
+        logger.info({ bestQuote: bestQuote }, 'bestQuote');
       } else {
         cosignerData = await getDefaultCosignerData(request, orderType, provider);
-        log.info({ cosignerData: cosignerData }, 'open order with default cosignerData');
+        logger.info({ cosignerData: cosignerData }, 'open order with default cosignerData');
       }
 
       const cosignedOrder = await createCosignedOrder(cosigner, request, cosignerData);
       // if no quote and creating open order, create random new quoteId
       const postedQuoteId = bestQuote?.quoteId ?? request.quoteId ?? request.requestId;
       try {
-        metric.putMetric(Metric.QUOTE_POST_ATTEMPT, 1, MetricLoggerUnit.Count);
+        metrics.increment(Metric.QUOTE_POST_ATTEMPT);
         const response = await orderServiceProvider.postOrder({
           order: cosignedOrder,
           signature: request.innerSig,
@@ -140,7 +143,7 @@ export class QuoteHandler extends APIGLambdaHandler<
           requestId: request.requestId,
         });
         if (response.statusCode == 200 || response.statusCode == 201) {
-          metric.putMetric(Metric.QUOTE_200, 1, MetricLoggerUnit.Count);
+          metrics.increment(Metric.QUOTE_200);
           // 200 and 201 (the latter also covers a post whose timeout was reconciled as
           // accepted) are the only confirmed posts, so this is the only place the
           // fade-breaker bookkeeping rows are written: the PostedOrders row and, for the
@@ -154,15 +157,14 @@ export class QuoteHandler extends APIGLambdaHandler<
             quote: bestQuote ?? undefined,
             quoteId: postedQuoteId,
             requestId: request.requestId,
-            log,
-            metric,
+            ctx,
           });
-          metric.putMetric(Metric.QUOTE_LATENCY, Date.now() - start, MetricLoggerUnit.Milliseconds);
+          metrics.histogram(Metric.QUOTE_LATENCY, Date.now() - start);
           const hardResponse = createHardQuoteResponse(request, cosignedOrder);
           if (!bestQuote) {
             // The RFQ responses are logged in getBestQuote()
             // we log the Open Orders here
-            log.info({
+            logger.info({
               eventType: RESPONSE_LOG_TYPE,
               body: {
                 ...hardResponse.toLog(),
@@ -176,11 +178,11 @@ export class QuoteHandler extends APIGLambdaHandler<
           };
         } else {
           const error = response as ErrorResponse;
-          log.error({ error: error }, 'Error posting order');
+          logger.error({ error: error }, 'Error posting order');
 
           // user error should not be alerted on
           if (error.detail != POST_ORDER_ERROR_REASON.INSUFFICIENT_FUNDS) {
-            metric.putMetric(Metric.QUOTE_POST_ERROR, 1, MetricLoggerUnit.Count);
+            metrics.increment(Metric.QUOTE_POST_ERROR);
           }
           // Only a 4xx from the order service is a genuine rejection of the
           // order. Anything else (timeouts, 5xx) is indeterminate — the order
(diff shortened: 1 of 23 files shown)


aayala-uniswap:
Read through this end to end. The change itself is safe and well tested — the EMF parity test replaying all 14 handler metrics against a real `MetricsLogger` is exactly the right check, and the blast radius is the two quote handlers plus the recorder. No objection to the approach.

One thing I'd like to fix before this lands, because it's cheap now and expensive after PR 2/3.

### The new interfaces don't match the monorepo's, and the PR says they do

The description says field names match the monorepo's ctx so handlers port without renames. The **field** names do (`logger`, `metrics`, `requestId`). The **method** shapes underneath don't. Compared against backend `main`:

| | this PR (`lib/observability`) | monorepo (`packages/lib/uni/interface.ts`, `packages/lib/observability/metrics.ts`) |
|---|---|---|
| Log call shape | bunyan style: `(fields, msg)` | `ILogger`: `(msg, ...extra)` — message first |
| Log levels | `trace` … `fatal` | `debug` … `fatal`, no `trace` |
| `child` | `child(bindings)` | `child()` (no args) |
| Metric methods | `increment` / `gauge` / `histogram` | `IMetrics`: `count` / `hist` / `gauge` / `set` / `timer` / `dist` |
| Metric return | sync `void` | `Promise<void>` |
| Tags | `MetricTags` map with reserved `status` / `reason` | `opts: { tags: string[], dimensions: Record<string,string> }` |

So every call site this PR rewrites — and everything PR 2 and PR 3 rewrite — gets rewritten again at port time. The seam is real, but it's shaped like bunyan + smart-order-router rather than like the thing it's meant to port onto.

### Suggestion

1. **Shape `Metrics` as the `IMetrics` subset we actually use**: `count(name, val?, opts?)`, `hist(name, val, opts?)`, `gauge(name, val, opts?)`, all `Promise<void>`. `EmfMetrics` does the same `putMetric` and returns a resolved promise; handlers keep calling without `await`, same as today.
2. **Shape `Logger` message-first like `ILogger`**, with a ~10-line adapter over the bunyan child that flips `(msg, fields)` → `(fields, msg)`. Bunyan stops satisfying the type structurally, but call sites are then written once in their final form.
3. **Drop `MetricReason` / `MetricTags` for now.** Nothing in this PR uses them, and the monorepo's tags are a string array, so they'd be discarded at port time anyway. Easy to add back when a call site needs one.

This PR touches four production files, so the reshape is small here. After PR 2 moves the whole quote path onto `ctx`, it isn't.

If you'd rather keep the current shape and eat a mechanical codemod at port time, I won't block on it — nothing here is incorrect. Just flagging that the cost is paid once now or three times later.

Minor / no action: the injector comment explaining why `setGlobalLogger` / `setGlobalMetric` still run is useful; keep it through PR 3. And the `putMetric`-throws-into-a-500 → log-and-drop change is a real improvement, worth a line in the release notes.

aayala-uniswap (dismissed):

alanhwu:
Thanks — reshaped as suggested, with three adjustments after checking against backend `main`:

- **`Metrics` is now the `IMetrics` subset we use: `count` / `timer` / `gauge`, `Promise<void>`, `opts?: Partial<MetricOptions>`** (`tags: string[]` + `dimensions`). `timer` rather than `hist` for latencies: the root CLAUDE.md says `timer()` is for durations, services call `timer` ~197× vs `hist` ~10×, and `hist` throws "Method not implemented" in both Lambda `IMetrics` implementations while `CloudWatchMetric.timer` maps to Milliseconds — which is exactly this repo's mapping.
- **Call sites `await` in the main flow and `void` in `catch`/`finally`**, per the monorepo's `no-floating-promises` + fire-and-forget rule, so they don't change again at port. `EmfMetrics` emits synchronously inside the async methods (tested), so a `void`-ed call lands before the handler returns.
- **`Logger` is message-first** (`debug`…`fatal`, `child()`, `setDefaultExtra`) over a ~40-line `BunyanLogger` adapter. The three analytics event lines pass `''` as the message: bunyan writes `"msg":""` for a fields-only call, so the records are byte-identical (`bunyan-logger.test.ts` asserts record equality), and the CW subscription filters key on `$.eventType` anyway.
- Dropped `MetricTags` / `MetricReason`; kept the monorepo's `MetricOptions` shape so `status:` / `reason:` string tags slot in when PR 2 moves the failure paths (root CLAUDE.md requires them on every operation).

Description updated to state what matches now (field names + method subset) and what still changes at port (metric names / service prefix / `.dist`, Datadog allowlists, vitest + `buildTestContext`). Same wire behaviour: 474 tests, 38 snapshots unmodified, synth differs only in the two quote Lambda assets.

aayala-uniswap (approved):

PR #508, opened 2026-09-22
fix(cron): drain stdout before the fade cron returns so its last log lines are not lost

alanhwu:
## What

The fade cron's last log lines are being truncated or dropped since the source flip (#504). The handler now awaits a stdout drain in a `finally` before returning.

## Why

After #504 the Redshift path runs **last**, as the shadow. Its ~160 KB result line is written seconds before the handler returns, followed by the shadow report. In Lambda `process.stdout` is a pipe, which Node writes asynchronously; when the runtime freezes the process the pipe is still draining. Observed on the first 14 post-flip runs: the Redshift line cut at exactly 65,536 or 131,072 bytes on 6 runs, and the shadow report (the run's last write) missing on 8 of 14 — including 2 where the big line landed but the report did not. The EMF metrics line was intact every time because Lambda writes `console.log` synchronously.

Decisions and metrics were never affected; only the log record was. But the shadow report is the comparison detail we watch, so this closes it before the few-day observation window is over.

## Fix

`flushStream(stream)` queues an empty write and resolves on its callback (i.e. once everything before it has been flushed), bounded at 2 s. The handler awaits `flushStdout()` in a `finally`, so failures flush too.

## Verification

- `yarn build` ✅ · `yarn test:unit` ✅ · `yarn lint` ✅ 0 errors
- Tests: `flushStream` resolves only after a pending 64 KiB write is consumed by a slow reader; resolves immediately with nothing pending; `flushStdout` resolves against the real stdout.
- No CDK change.

## Follow-up worth considering

The two ~170 KB-per-run lines (`getFillersFadeStats` rows and the Redshift result) are ~50 MB/day of log volume in a never-expiring group; trimming them is a separate, cheap cost win.


alanhwu:
Closing: not needed for the observation window (metrics are unaffected; the report still lands on a share of runs), and the Redshift path that writes the large late line is slated for removal once the new source is confirmed.

